PR #2, opened 2020-09-16
8253211: Make sure jvm does not crash with Arm SVE and Vector API

Jamsauce:
Add those vector api specific OPs to unsupported list for SVE.

Currently we have not implemented all Arm SVE code generation for
Vector API specific nodes. To make sure hotspot does not crash with bad
AD file (as NEON has implemented them), we simply add those OPs to
unsupported op list.

This patch also fixed one missing VectorA handling in previous merge.


Jamsauce:
Used the wrong ID

PR #19, opened 2020-10-29
8255437: [vector] jdk/incubator/vector/SelectFromByteMaxVectorTests fails with Arm SVE 2048-bits

XiaohongGong:
The tests are used to test the Vector API "selectFrom". It uses index values stored in the lanes of this vector which is a byte vector for byte type. So its valid lane elements should be 0 ~ 127 for byte.

However, for SVE 2048-bits, the max vector length for byte is 256, which exceeds the byte boundary. Indexes greater than 127 are converted to the negative values which are invalid for the API.

To avoid this, it's better to guarantee the elements in the byte vector to the valid values if they are prepared to be used as vector indexes.

See more details from:
https://mail.openjdk.java.net/pipermail/panama-dev/2020-October/011216.html

---------
### Progress

### Testing

|     | Linux x32 | Linux x64 | Windows x64 | macOS x64 |
| --- | ----- | ----- | ----- | ----- |
| Build | ✔️ (1/1 passed) | ✔️ (5/5 passed) | ❌ (2/2 failed) | ✔️ (2/2 passed) |
| Test (tier1) |    |  ✔️ (9/9 passed) |    |  ✔️ (9/9 passed) |

**Failed test tasks**
- [Windows x64 (build debug)](https://github.com/XiaohongGong/panama-vector/runs/1331094421)
- [Windows x64 (build release)](https://github.com/XiaohongGong/panama-vector/runs/1331094392)

### Issue
 * [JDK-8255437](https://bugs.openjdk.java.net/browse/JDK-8255437): [vector] jdk/incubator/vector/SelectFromByteMaxVectorTests fails with Arm SVE 2048-bits

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to ccf933d1ad33f33f2e3daea71612211f7d662544

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/19/head:pull/19`
`$ git checkout pull/19`

### DIFF
--- a/test/jdk/jdk/incubator/vector/templates/Unit-header.template
+++ b/test/jdk/jdk/incubator/vector/templates/Unit-header.template
@@ -1184,8 +1184,13 @@ public class $vectorteststype$ extends AbstractVectorTest {
     static final List<BiFunction<Integer,Integer,$type$[]>> $TYPE$_SHUFFLE_GENERATORS = List.of(
             withToStringBi("shuffle[random]", (Integer l, Integer m) -> {
                 $type$[] a = new $type$[l];
+#if[ByteMax]
+                int upper = Math.min(Byte.MAX_VALUE + 1, m);
+#else[ByteMax]
+                int upper = m;
+#end[ByteMax]
                 for (int i = 0; i < 1; i++) {
-                    a[i] = ($type$)RAND.nextInt(m);
+                    a[i] = ($type$)RAND.nextInt(upper);
                 }
                 return a;
             })
--- a/test/jdk/jdk/incubator/vector/Byte128VectorTests.java
+++ b/test/jdk/jdk/incubator/vector/Byte128VectorTests.java
@@ -980,8 +980,9 @@ public class Byte128VectorTests extends AbstractVectorTest {
     static final List<BiFunction<Integer,Integer,byte[]>> BYTE_SHUFFLE_GENERATORS = List.of(
             withToStringBi("shuffle[random]", (Integer l, Integer m) -> {
                 byte[] a = new byte[l];
+                int upper = m;
                 for (int i = 0; i < 1; i++) {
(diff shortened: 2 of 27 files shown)


test/jdk/jdk/incubator/vector/templates/Unit-header.template
```diff
@@ -1184,8 +1184,15 @@ public class $vectorteststype$ extends AbstractVectorTest {
     static final List<BiFunction<Integer,Integer,$type$[]>> $TYPE$_SHUFFLE_GENERATORS = List.of(
             withToStringBi("shuffle[random]", (Integer l, Integer m) -> {
                 $type$[] a = new $type$[l];
+#if[ByteMax]
```
PaulSandoz, line 1187: You could probably simplify this a bit by rename argument `m` to `upper` and updating it for `ByteMax`, thereby the loop code remains the same.

PaulSandoz (approved):

test/jdk/jdk/incubator/vector/templates/Unit-header.template
XiaohongGong: Hi @PaulSandoz thanks for looking at this PR.  I'v updated the patch to make sure the loop code remain the same. Could you please to take a look at it again? Thanks!

PaulSandoz:
Latest changes are fine, but you can just rename the `m` argument of the lambda expression to `upper` and make it even simpler :-) up to you, no need for another round of review.

XiaohongGong:
@PaulSandoz  Thanks for looking at it again and thanks for your advice! I'd like to keep it as it is. Thanks again!

nsjian:
/integrate

Note: a force-push will review history.

nsjian:
/sponsor

Note: a force-push will break review history.

PR #22, opened 2020-11-16
Load/store short vectors from/to char[]

PaulSandoz:
Methods to load/store a `ShortVector` from/to `char[]`. These methods may be useful for encoding/decoding characters.

Unclear if aliasing of `char[]` and `short[]` causes issues for the intrinsic or for Hotspot code gen.

Testing-wise, it was easier to explicitly clone and reify the relevant tests rather than trying to generalize the template and its template arguments to distinguish between the vector type and the array type.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/22/head:pull/22`
`$ git checkout pull/22`

### DIFF
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
@@ -805,6 +805,7 @@ final class Byte128Vector extends ByteVector {
         return super.fromArray0Template(a, offset);  // specialize
     }
 
+
     @ForceInline
     @Override
     final
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte256Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte256Vector.java
@@ -837,6 +837,7 @@ final class Byte256Vector extends ByteVector {
         return super.fromArray0Template(a, offset);  // specialize
     }
 
+
     @ForceInline
     @Override
     final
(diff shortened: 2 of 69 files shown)


luhenry:
Hey @PaulSandoz, thank you for doing that! It works for https://gist.github.com/luhenry/dc84c661997d5bc1cd51c2088d234727 so far. However, as I'm looking at expanding for CharBuffer as well, I'm seeing that there is a `fromByteBuffer`/`intoByteBuffer` method, but not a `fromCharBuffer`/`intoCharBuffer`.

PaulSandoz:
I was hoping to avoid adding support for `CharBuffer` load/store. We have so far avoided adding support for `IntBuffer` etc. When `MemorySegment` exits incubation we may even remove the `ByteBuffer` methods (since segment can wrap them).

Can you operate on `char[]` instead for now? I see your code requires that the `dst` buffer has a backing `char[]` array, although perhaps you intend to support direct `CharBuffer` too?

luhenry:
> Can you operate on char[] instead for now? I see your code requires that the dst buffer has a backing char[] array, although perhaps you intend to support direct CharBuffer too?

I absolutely can only rely on a CharBuffer backed by an array only. Adding support for CharBuffer without an array is not required to make it work and to use it as an experiment, it would just make it more feature complete. I was just raising the question as I wasn't sure if it was an oversight or not, and from your explanation, it isn't.

PaulSandoz:
Unsure, we might if the algorithms demand that?

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
```diff
@@ -805,6 +805,7 @@ ByteVector fromArray0(byte[] a, int offset) {
         return super.fromArray0Template(a, offset);  // specialize
     }
 
+
```
DamonFool, line 808: Some files were just modified by adding an empty newline.
Why?

PaulSandoz: It's an unfortunate side-effect of generating files from a template , where in some cases the generation is partial.
In this case generating from `X-VectorBits.java.template` for code specific to the `short` element type.

sviswa7 (approved):

PR #26, opened 2020-11-24
Add utf8 decoding benchmarks

luhenry:
Following discussions on the mailing list, I'm submitting three benchmarks around UTF-8 decoding:
 - decode: uses a while-loop based implementation currently in use in the JDK
 - decodeVector: uses a lookup table with vector operations for 1-3 bytes characters
 - decodeVectorASCII: uses a simple vector operation to accelerate parsing ASCII-only characters

We don't observe the expected speedups with either decodeVector and decodeVectorASCII, so these are, I think, good test cases to further develop the Vector API.

---------
### Progress

### Testing

|     | Linux aarch64 | Linux arm | Linux ppc64le | Linux s390x | Linux x64 | Linux x86 | Windows x64 | macOS x64 |
| --- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- |
| Build | ✔️ (1/1 passed) | ✔️ (1/1 passed) | ✔️ (1/1 passed) | ✔️ (1/1 passed) | ✔️ (6/6 passed) | ✔️ (2/2 passed) | ❌ (2/2 failed) | ✔️ (2/2 passed) |
| Test (tier1) |    |     |     |     |  ✔️ (9/9 passed) | ✔️ (9/9 passed) |    |  ✔️ (9/9 passed) |

**Failed test tasks**
- [Windows x64 (build debug)](https://github.com/luhenry/panama-vector/runs/1458281921)
- [Windows x64 (build release)](https://github.com/luhenry/panama-vector/runs/1458281893)

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/26/head:pull/26`
`$ git checkout pull/26`

### DIFF
--- a/test/jdk/jdk/incubator/vector/benchmark/src/main/java/benchmark/utf8/DecodeBench.java
+++ b/test/jdk/jdk/incubator/vector/benchmark/src/main/java/benchmark/utf8/DecodeBench.java
@@ -56,9 +56,9 @@ public class DecodeBench {
     private String in;
     private String out;
 
-    private static final VectorSpecies<Byte>  B64  = ByteVector.SPECIES_64;
     private static final VectorSpecies<Byte>  B128 = ByteVector.SPECIES_128;
     private static final VectorSpecies<Short> S128 = ShortVector.SPECIES_128;
+    private static final VectorSpecies<Short> S256 = ShortVector.SPECIES_256;
 
     private static final HashMap<Long, DecoderLutEntry> lutTable = new HashMap<Long, DecoderLutEntry>();
 
@@ -287,7 +287,7 @@ public class DecodeBench {
     }
 
     @Benchmark
-    public void decode() {
+    public void decodeScalar() {
         decodeArrayLoop(src, dst);
     }
 
@@ -369,15 +369,13 @@ public class DecodeBench {
         int dl = dst.arrayOffset() + dst.limit();
 
         // Vectorized loop
-        while (sp + B64.length() - 1 < sl && dp + S128.length() - 1 < dl) {
-            var bytes = ByteVector.fromArray(B64, sa, sp);
+        for (; sp <= sl - B128.length() && dp <= dl - S256.length(); sp += B128.length(), dp += S256.length()) {
+            var bytes = ByteVector.fromArray(B128, sa, sp);
 
-            if (bytes.compare(VectorOperators.LT, (byte)0x00).anyTrue())
+            if (bytes.compare(VectorOperators.LT, (byte) 0x00).anyTrue())
                 break;
 
-            ((ShortVector)bytes.convertShape(VectorOperators.B2S, S128, 0)).intoCharArray(da, dp);
-            sp += B64.length();
-            dp += S128.length();
+            ((ShortVector) bytes.convertShape(VectorOperators.B2S, S256, 0)).intoCharArray(da, dp);
         }
 
         updatePositions(src, sp, dst, dp);


rose00:
Yes, let's figure out how to make LUT-based algorithms really "sing".

I wonder if there's any benefit to intrinsifying some or all of the steps between deriving a syndrome number and applying the resulting shuffle(s).  In this example the steps are:  Do a `compare`, convert the comparison to a scalar bit mask (syndrome number), use it as a `get` key on a Java object, make some more indirections, you grab a shuffle vector, and use it to steer the original data.  That's a lot of stuff for the JIT to "see through".

luhenry:
> I wonder if there's any benefit to intrinsifying some or all of the steps between deriving a syndrome number and applying the corresponding selected shuffle(s). In this example the steps are: Do a compare, convert the comparison to a scalar bit mask (syndrome number), use it as a get key on a Java object, make some more indirections, grab a shuffle vector, and finally use it to steer the original data. There's also bits of control flow interspersed. 

Yes, you summarize the steps well. Other than the validation step, all of them are fairly straightforward. Offloading as much as possible ahead-of-time (the lookup table) really is what gets you the speed (in theory). However, as discussed on the mailing list, this has the disadvantage of computing all this information at startup.

What's interesting is that the same algorithm implemented in C++ yields significant improvements, but we're not getting similar gains with the Vector API. I'm learning my way through why the JIT isn't able to generate the right code, and I'll come back with my results to the mailing list. In the meantime, would merging the PR with this current implementation be good for you?

> That's a lot of stuff for the JIT to "see through".

Do you mean to "see through" in this case of using the Vector API? To try to help the JIT, I've tried storing in DecoderLutEntry only the byte[] and short[] backing the decoding and validation vectors, and load the vectors from that. However, I didn't see any performance improvement with that, and I even observed a non-negligible regression.

iwanowww (commented):
Looks good.

sviswa7:
@luhenry Looks good to me as well. Please do create a separate branch (other than vectorIntrinsics) as indicated above by OpenJDK Skara bot and resubmit the pull request from the new branch.

PaulSandoz:
This benchmark has identified a few issues:

1. Operations on species of byte/64 are not intrinsic.
2. `ShortVector` from/to char[] are not intrinsic (i suspected this to be the case).

For now:

1) can be unblocked by focusing on byte/128 and short/256.
2) can be unblocked with the following patch (so it appears, but needs more detailed review) or using `short[]`.

```
index db7b69a9137..399130b0e45 100644
@@ -624,7 +624,10 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
   // Handle loading masks.
   // If there is no consistency between array and vector element types, it must be special byte array case or loading masks
   if (arr_type != NULL && !using_byte_array && elem_bt != arr_type->elem()->array_element_basic_type() && !is_mask) {
-    return false;
+    if (elem_bt == T_SHORT && arr_type->elem()->array_element_basic_type() == T_CHAR) {
+    } else {
+      return false;
+    }
   }
```

In general, to found the causes of issues i recommend extracting out vector sub-expressions and placing in separate benchmarks. It's easier to analyze the code that is generated.

The benchmark is also storing vectors/shuffles on the heap. Instead i recommend storing such data in compatible arrays, then loading into vector instances held in local variables.

PaulSandoz:
Clarifications on byte/64 operations are not intrinsic. I was just focusing on the `decodeVectorASCII` benchmark. I think it is specifically due to the mask operations.

luhenry:
> @luhenry Looks good to me as well. Please do create a separate branch (other than vectorIntrinsics) as indicated above by OpenJDK Skara bot and resubmit the pull request from the new branch.

@sviswa7 the only issue it introduces is that, once this PR is merged, I'll need to force-push to my [luhenry/panama-vector:vectorIntrinsics](https://github.com/luhenry/panama-vector/commits/vectorIntrinsics) branch as it will have diverged from [openjdk/panama-vector:vectorIntrinsics](https://github.com/openjdk/panama-vector/commits/vectorIntrinsics) branch. But that is something I'm used to working with, and thus shouldn't be a blocker for this PR.

luhenry:
@PaulSandoz thank you for your detailed explanation. 

> In general, to found the causes of issues i recommend extracting out vector sub-expressions and placing in separate benchmarks. It's easier to analyze the code that is generated.

That should work for the ASCII case easily, but we would need to disable the verification in `DecodeBench.tearDownInvocation`. I am taking that this is acceptable when benchmarking and debugging local changes.

> The benchmark is also storing vectors/shuffles on the heap. Instead i recommend storing such data in compatible arrays, then loading into vector instances held in local variables.

Let me change that locally and see what numbers I get with that.

sviswa7 (approved):

PR #29, opened 2020-12-16
8258072: Add Reinterpret nodes implementation for Arm SVE

Wanghuang-Huawei:
Add reinterpret nodes implementation for Arm SVE. We also adjust the implementation of Matcher::min_vector_size in the same patch.

---------
### Progress

### Issue
 * [JDK-8258072](https://bugs.openjdk.java.net/browse/JDK-8258072): Add Reinterpret nodes implementation for Arm SVE

### Contributors
 * Xuejin He `<hexuejin2@huawei.com>`

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/29/head:pull/29`
`$ git checkout pull/29`


src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -250,7 +249,7 @@ definitions %{
 
 // All SVE instructions
 
-// vector load/store
+// vector load/store/reinterpret
```
XiaohongGong, line 252: Can you please add a separate comment for "reinterpret" before the related rules?

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -340,7 +339,38 @@ instruct storeV_partial(vReg src, vmemA mem, pRegGov pTmp, iRegINoSp tmp1,
   %}
   ins_pipe(pipe_slow);
 %}
+instruct reinterpret(vReg dst) %{
+  predicate(UseSVE > 0 && n->as_Vector()->length_in_bytes() >= 16 &&
+            n->in(1)->bottom_type()->is_vect()->length_in_bytes() >= 16 &&
```
XiaohongGong, line 344: This check in line-344 is redundant as the pre and subsequent check can guarantee this.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -340,7 +339,38 @@ instruct storeV_partial(vReg src, vmemA mem, pRegGov pTmp, iRegINoSp tmp1,
   %}
   ins_pipe(pipe_slow);
 %}
+instruct reinterpret(vReg dst) %{
```
XiaohongGong, line 342: Could you please try to add one more line between these different rules?

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -2498,7 +2498,14 @@ const int Matcher::min_vector_size(const BasicType bt) {
       // To support vector api load/store mask.
       return MaxVectorSize / 8;
     }
-    return max_size;
+    // Limit the vector size to 8 bytes
+    int size = 32 / type2aelembytes(bt);
+    if (bt == T_BYTE) {
+      // To support vector api shuffle/rearrange.
+      size = 8;
```
XiaohongGong, line 2505: Any reason that the min vector size for `"shuffle/rearrange"` is `8` while not `32/64 = 4` ?

src/hotspot/cpu/aarch64/aarch64_sve.ad
Wanghuang-Huawei: This comment is copied from neon implementation. We will add a  separate comment for "reinterpret"  in revised patch.

src/hotspot/cpu/aarch64/aarch64_sve.ad
Wanghuang-Huawei: ok. Thanks.

src/hotspot/cpu/aarch64/aarch64_sve.ad
Wanghuang-Huawei: Sure

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -1833,7 +1833,7 @@ bool LibraryCallKit::inline_vector_convert() {
     } else if (num_elem_from > num_elem_to) {
       // Since number elements from input is larger than output, simply reduce size of input (we are supposed to
       // drop top elements anyway).
-      int num_elem_for_resize = MAX2(num_elem_to, Matcher::min_vector_size(elem_bt_to));
+      int num_elem_for_resize = MAX2(num_elem_to, Matcher::min_vector_size(elem_bt_from));
```
iwanowww, line 1836: Looks like a bug to me. 
Can you elaborate, please, how it manifests and does it cause any misbehavior? 
I'm concerned that it should be fixed separately upstream.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -340,7 +339,38 @@ instruct storeV_partial(vReg src, vmemA mem, pRegGov pTmp, iRegINoSp tmp1,
...
+  ins_cost(SVE_COST);
+  format %{ " # reinterpret $dst,$src" %}
+  ins_encode %{
+    // If register is the same, then move is not needed.
+    if (as_FloatRegister($dst$$reg) != as_FloatRegister($src$$reg)) {
+      __ sve_orr(as_FloatRegister($dst$$reg),
+             as_FloatRegister($src$$reg),
+             as_FloatRegister($src$$reg));
+    }
+  %}
```
XiaohongGong, line 373: Did you consider the resize case when the dst vector length in bytes is higher than the src vector length in bytes? I understand that for the parts that the length is `LE` to the src vector length, the values should be moved from the src values, otherwise the dst values are expected 0. Do you think it needs to use the SVE predicate feature to control the action here? 

For an example:
```
   1. byte[] a = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
   2. byte[] b = new byte[8];
       b = reinterpret(a) ;    // b = [1, 2, 3, 4, 5, 6, 7, 8] 
   3. byte[] c = new byte[16];
       c = reinterpret(b);     // c = [1, 2, 3, 4, 5, 6, 7, 8, 0, 0, 0, 0, 0, 0, 0, 0]
```
If no vector length info is referenced here, I'm wondering whether the final result is:
```
    c = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```

src/hotspot/share/opto/vectorIntrinsics.cpp
Wanghuang-Huawei: Yes, we think it is a bug.  I reported a bug : https://bugs.openjdk.java.net/browse/JDK-8258989

PR #37, opened 2021-02-09
8261108: Add cast nodes from integer types to float types implementation for Arm SVE

Wanghuang-Huawei:
Add the backend codegen for nodes like castb2f / castb2d / casts2f / casts2d / casti2f / casti2d / castl2f / castl2d.

* add `sve_scvtf`  ( and `sve_ucvtf`) encoding
* add mock for these instructions
* add `VECTOR_CAST_I2F_EXTEND[N]` and `VECTOR_CAST_I2F`

Thank you for your review.

---------
### Progress

### Issue
 * [JDK-8261108](https://bugs.openjdk.java.net/browse/JDK-8261108): Add cast nodes from integer types to float types implementation for Arm SVE

### Reviewers

### Contributors
 * He Xuejin `<hexuejin2@huawei.com>`

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/37/head:pull/37`
`$ git checkout pull/37`

### DIFF
--- a/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
@@ -3356,7 +3356,7 @@ public:
 private:
 
   void encode_cvtf_T(SIMD_RegVariant T_dst, SIMD_RegVariant T_src,
-		      int& opc, int& opc2) {
+                     int& opc, int& opc2) {
     assert(T_src != B && T_dst != B &&
            T_src != Q && T_dst != Q, "invalid register variant");
     if (T_dst != D) {


src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3353,6 +3353,38 @@ void mvnw(Register Rd, Register Rm,
   INSN(sve_whilelsw, 0b111, 0);
 #undef INSN
 
+  void encode_cvtf_T (SIMD_RegVariant& T_dst, SIMD_RegVariant& T_src) {
```
nsjian, line 3356: It's better to be a non-public function.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
```diff
@@ -1587,3 +1587,108 @@ instruct vcvt$1to$2`'(vReg dst, vReg src, vReg tmp)
...
+  ins_encode %{
+    __ sve_$3(as_FloatRegister($dst$$reg), __ $4, as_FloatRegister($src$$reg));
+    __ sve_$3(as_FloatRegister($dst$$reg), __ $5, as_FloatRegister($dst$$reg));
+    __ sve_$3(as_FloatRegister($dst$$reg), __ $6, as_FloatRegister($dst$$reg));
+    __ sve_$7(as_FloatRegister($dst$$reg), __ $6, ptrue, as_FloatRegister($dst$$reg), __ $6);
+  %}
+  ins_pipe(pipe_slow);
+%}')dnl
+dnl             $1 $2 $3       $4 $5 $6 $7
+VECTOR_CAST_I2F_EXTEND3(B, D, sunpklo, H, S, D, scvtf)
```
nsjian, line 1634: Can you please align the comment $1...$7 to the arguments?

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
```diff
@@ -1587,3 +1587,108 @@ instruct vcvt$1to$2`'(vReg dst, vReg src, vReg tmp)
...
+%{
+  predicate(UseSVE > 0 && n->bottom_type()->is_vect()->length_in_bytes() >= 16 &&
+            n->bottom_type()->is_vect()->element_basic_type() == T_`'TYPE2DATATYPE($2));
+  match(Set dst (VectorCast$1`'2X src));
+  effect(TEMP_DEF dst);
+  ins_cost(3 * SVE_COST);
+  format %{ "sve_$3  $dst, $4, $src\n\t"
+            "sve_$3  $dst, $5, $dst\n\t"
+            "sve_$3  $dst, $6, $dst\n\t"
+            "sve_$6  $dst, $7, $dst, $7\t# convert $1 to $2 vector" %}
```
nsjian, line 1624: sve_$7 instead?

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3353,6 +3353,38 @@ void mvnw(Register Rd, Register Rm,
...
+    starti;                                                             \
+    f(0b01100101, 31, 24), f(0b010, 21, 19), f(0b101, 15, 13);          \
+    f(sign, 16);                                                        \
+    pgrf(Pg, 10), rf(Zn, 5), rf(Zd, 0);                                 \
+    encode_cvtf_T(T_dst, T_src);                                        \
+    f(T_dst, 23, 22), f(T_src, 18, 17);                                 \
+  }
+
+  INSN(sve_scvtf, 0b0);
+  INSN(sve_ucvtf, 0b1);
```
nsjian, line 3385: Is there any usage for ucvtf?

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -2684,3 +2684,135 @@ instruct vcvtLtoB(vReg dst, vReg src, vReg tmp)
...
+  ins_pipe(pipe_slow);
+%}
+
+
+instruct vcvtBtoD(vReg dst, vReg src)
+%{
+  predicate(UseSVE > 0 && n->bottom_type()->is_vect()->length_in_bytes() >= 16 &&
+            n->bottom_type()->is_vect()->element_basic_type() == T_DOUBLE);
+  match(Set dst (VectorCastB2X src));
+  effect(TEMP_DEF dst);
```
nsjian, line 2728: This effect looks unnecessary to me.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
Wanghuang-Huawei: Thank you for your review. I will fix this.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: Yes, I will fix it.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
Wanghuang-Huawei: It's my fault. I'll change that.

src/hotspot/cpu/aarch64/aarch64_sve.ad
Wanghuang-Huawei: OK. I will remove that.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: Just like `sve_uzp2`.It seems that  `ucvtf` and `scvtf`  are in the same encoding group.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3353,6 +3353,42 @@ void mvnw(Register Rd, Register Rm,
...
+// SVE convert integer to floating-point (predicated)
+#define INSN(NAME, sign)                                                \
+  void NAME(FloatRegister Zd, SIMD_RegVariant T_dst, PRegister Pg,      \
+            FloatRegister Zn, SIMD_RegVariant T_src) {                  \
+    starti;                                                             \
+    f(0b01100101, 31, 24), f(0b010, 21, 19), f(0b101, 15, 13);          \
+    f(sign, 16);                                                        \
+    pgrf(Pg, 10), rf(Zn, 5), rf(Zd, 0);                                 \
+    encode_cvtf_T(T_dst, T_src);                                        \
+    f(T_dst, 23, 22), f(T_src, 18, 17);                                 \
```
nsjian, line 3385: I would suggest to fill the bits in order, i.e. move f(T_dst, 23, 22) right after (... 31, 24), and then (21, 19), (18, 17)...

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3353,6 +3353,42 @@ void mvnw(Register Rd, Register Rm,
...
+      assert(T_dst <= T_src, "invalid register variant");
+    } else {
+      assert(T_src != H, "invalid register variant");
+    }
+    if (T_dst == S && T_src == D) {
+      T_dst = D;
+      T_src = S;
+    } else if (T_dst == D && T_src == S) {
+      T_dst = D;
+      T_src = B;
```
nsjian, line 3371: This looks confusing. I understand that it's just 0b00, but using B here is really confusing. Maybe using 0b00 and 0b11 directly to align with Arm ARM opc/opc2?

Wanghuang-Huawei: Arm uses confusing encoding inst. If we use 0b00 / 0b11 ,  we can not get a concise macroassmbler here ( for example. other developer can not use this macro). In fact, `gcc` uses switch..case to convert macro assmbler into Arm's funny encoding.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: `T_dst` and `T_src` will be changed after  `encode_cvtf_T`, so they cannot move there.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: If we use an `encoding function `(like `switch..case` in `gcc`) instead of `convert function`(`encode_cvtf_T`),  source codes here will be a little complicated. :-)

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
nsjian: I think it's fine to assign the values at the beginning. And maybe you can add more test cases to check different combinations.

Wanghuang-Huawei: I have added some comments here to explain codes here.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
XiaohongGong: I agree with @nsjian that swapping "T_dst" and "T_src" is confusing although this can generate the same final result.  As I understand that this function is used to generate two opcs from the dst and src type, I'd like not using the swap logic. Thanks!

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
XiaohongGong: Agree with @nsjian that filling the bits in order which can make the code more clean. Thanks!

Wanghuang-Huawei: OK. I will change this order.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -2684,3 +2684,135 @@ instruct vcvtLtoB(vReg dst, vReg src, vReg tmp)
...
+  match(Set dst (VectorCastL2X src));
+  effect(TEMP_DEF dst, TEMP tmp);
+  ins_cost(3 * SVE_COST);
+  format %{ "sve_scvtf  $dst, S, $src, D\n\t"
+            "sve_dup  $tmp, S, 0\n\t"
+            "sve_uzp1  $dst, S, $dst, $tmp\t# convert L to F vector" %}
+  ins_encode %{
+    __ sve_scvtf(as_FloatRegister($dst$$reg), __ S, ptrue, as_FloatRegister($src$$reg), __ D);
+    __ sve_dup(as_FloatRegister($tmp$$reg), __ S, 0);
+    __ sve_uzp1(as_FloatRegister($dst$$reg), __ S, as_FloatRegister($dst$$reg), as_FloatRegister($tmp$$reg));
```
XiaohongGong, line 2756: This looks an issue to me. From the encoding, the `"__ sve_scvtf(as_FloatRegister($dst$$reg), __ S, ptrue, as_FloatRegister($src$$reg), __ D);"` can directly converts 64-bits to single-precision. Why do you need the followed `sve_uzp1` instruction? I guess you want convert "long" to "double" and then to "float" ? If so, the first `scvtf` instruction should be `"__ sve_scvtf(as_FloatRegister($dst$$reg), __ D, ptrue, as_FloatRegister($src$$reg), __ D);"`, and then encoding an `fcvt` which converts `"double"` to `"float"`. Please correct me if I'm wrong. Thanks!

XiaohongGong:
BTW, there is the floating-point conversions that are missing for SVE. Is it in you plan in a short term? Thanks!

Wanghuang-Huawei:
> BTW, there is the floating-point conversions that are missing for SVE. Is it in you plan in a short term? Thanks!
Yes. I have done that.  After this patch, I will commit that.

XiaohongGong:
> > BTW, there is the floating-point conversions that are missing for SVE. Is it in you plan in a short term? Thanks!
> > Yes. I have done that.  After this patch, I will commit that.

That's great! Thanks so much!

nsjian (changes requested):
Thanks for the work! I think you haven't addressed all the review comments. But I believe those are trivial for now, and I will do the code refactoring to address them when we have full SVE backend support.

Before integration, could you please fix the jcheck error?

Wanghuang-Huawei:
> Thanks for the work! I think you haven't addressed all the review comments. But I believe those are trivial for now, and I will do the code refactoring to address them when we have full SVE backend support.
> 
> Before integration, could you please fix the jcheck error?

It's my fault. I have fixed that.

nsjian (approved):

Wanghuang-Huawei:
Contributor Wang Huang <whuang@openjdk.org>

PR #38, opened 2021-02-10
8261394: [vector] Crash with "assert(Matcher::vector_size_supported(elem_bt, length)) failed: length in range"

XiaohongGong:
The crash is introduced by [1] and happens when the compiler is making an unspported vector type:
```
  assert(Matcher::vector_size_supported(elem_bt, length)) failed: length in range
```
Before vectorization for each vector intrinsic, the hotspot will check whether current arch supports the vector operation from several aspects like the backend implementation, the vector type and length. Once one of them is not matched, the compiler will stop the vectorization and go back to the default java implementation.

The changes in [1] missed the check for "Op_CallLeafVector". I think the double 64-bits vector is not supported to be vectorized. However, due to the missing check, the compiler continues the progress and then the crash happens.

This patch fixes it by making sure the check contains all opcodes.

[1] https://bugs.openjdk.java.net/browse/JDK-8261267

---------
### Progress

### Issue
 * [JDK-8261394](https://bugs.openjdk.java.net/browse/JDK-8261394): [vector] Crash with "assert(Matcher::vector_size_supported(elem_bt, length)) failed: length in range"

### Reviewers
 * [Ningsheng Jian](https://openjdk.java.net/census#njian) (@nsjian - Committer) ⚠️ Review applies to c7fb927616638260b3826ab3507ef4934d1ab64a
 * [Jie Fu](https://openjdk.java.net/census#jiefu) (@DamonFool - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/38/head:pull/38`
`$ git checkout pull/38`

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -253,11 +253,18 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
       }
       return false;
     }
+    if (!Matcher::vector_size_supported(elem_bt, num_elem)) {
+      if (C->print_intrinsics()) {
+        tty->print_cr("  ** vector size (vlen=%d, etype=%s) is not supported",
+                      num_elem, type2name(elem_bt));
+      }
+      return false;
+    }
   }
 
   // TODO When mask usage is supported, VecMaskNotUsed needs to be VecMaskUseLoad.
-  int opcode = sopc != 0 ? sopc : Op_CallLeafVector;
-  if (!arch_supports_vector(opcode, num_elem, elem_bt, is_vector_mask(vbox_klass) ? VecMaskUseAll : VecMaskNotUsed)) {
+  if ((sopc != 0) &&
+      !arch_supports_vector(sopc, num_elem, elem_bt, is_vector_mask(vbox_klass) ? VecMaskUseAll : VecMaskNotUsed)) {
     if (C->print_intrinsics()) {
       tty->print_cr("  ** not supported: arity=%d opc=%d vlen=%d etype=%s ismask=%d",
                     n, sopc, num_elem, type2name(elem_bt),


XiaohongGong:
Hi, could you anyone please take a look at this PR? Thanks so much!

nsjian (approved):
Looks good to me, but perhaps @sviswa7 may take a look as well?

DamonFool:
For ACOSDouble64VectorTests, I think @sviswa7 would like to generate code like this:

```
098     B6: #   out( B14 B7 ) <- in( B5 )  Freq: 0.999977
098     # checkcastPP of RBX
098     movl    R11, [RBX + #12 (8-bit)]        # compressed ptr ! Field: jdk/internal/vm/vector/VectorSupport$VectorPayload.payload (constant)
09c     load_vector XMM0,[R12 + R11 << 3 + #16] (compressed oop addressing)
0a3     call_leaf,vector  vector_acos_double64
        No JVM State Info
        #
```

However, the fix will generate code like this.

```
095     B6: #   out( B18 B7 ) <- in( B5 )  Freq: 0.999977
095     movq    R9, RAX # spill
098     # checkcastPP of R9
098     movq    RDI, jdk/incubator/vector/DoubleVector$$Lambda$27+0x0000000801052e28:exact *    # ptr
0a2     movl    RSI, #108       # int
0a7     movq    RDX, java/lang/Class:exact *    # ptr
0b1     movq    RCX, java/lang/Class:exact *    # ptr
0bb     movl    R8, #1  # int
        nop     # 3 bytes pad for loops and calls
0c4     call,static  jdk.internal.vm.vector.VectorSupport::unaryOp
        # jdk.incubator.vector.DoubleVector::lanewiseTemplate @ bci:71 (line 554) L[0]=_ L[1]=_ L[2]=_
        # jdk.incubator.vector.Double64Vector::lanewise @ bci:2 (line 273) L[0]=_ L[1]=_
        # jdk.incubator.vector.Double64Vector::lanewise @ bci:2 (line 41) L[0]=_ L[1]=_
        # Test::ACOSDouble64VectorTests @ bci:15 (line 10) L[0]=_
        # OopMap {rbp=Oop off=204/0xcc}
```

Maybe, it's OK to call vector_acos_double64 for Double64Vector on x86.
Thanks.

sviswa7:
@DamonFool is correct, with this change the vector stubs are not being called at all.
@XiaohongGong @nsjian The correct fix would be as follows:
index 056c415..b0ac90f 100644
@@ -253,6 +253,9 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
       }
       return false;
     }
+    if (!Matcher::vector_size_supported(elem_bt, num_elem)) {
+      return false;
+    }
   }

   // TODO When mask usage is supported, VecMaskNotUsed needs to be VecMaskUseLoad.

DamonFool:
> @DamonFool is correct, with this change the vector stubs are not being called at all.
> @XiaohongGong @nsjian The correct fix would be as follows:
> diff --git a/src/hotspot/share/opto/vectorIntrinsics.cpp b/src/hotspot/share/opto/vectorIntrinsics.cpp
> index 056c415..b0ac90f 100644
> --- a/src/hotspot/share/opto/vectorIntrinsics.cpp
> +++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
> @@ -253,6 +253,9 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
> }
> return false;
> }
> 
> * if (!Matcher::vector_size_supported(elem_bt, num_elem)) {
> * ```
>    return false;
>   ```
> * }
>   }
> 
> // TODO When mask usage is supported, VecMaskNotUsed needs to be VecMaskUseLoad.

This change doesn't call vector_acos_double64 for Double64Vector on x86 either.
But it seems reasonable and better to understand.

Let's push it to fix the bug ASAP.
And I'll check whether we can generate vector_acos_double64 for Double64Vector later.
Thanks.

XiaohongGong:
Hi @sviswa7 @DamonFool, the suggested solution makes sense to me. Thanks for looking at this change. I will update the patch ASAP.

DamonFool (approved):
Looks good to me.
Thanks for fixing it.

sviswa7 (approved):

XiaohongGong:
Thanks for the review @DamonFool @sviswa7 @nsjian

PR #40, opened 2021-02-25
RFC: Vector API masking support proposal for Arm SVE

XiaohongGong:
Please help to review this proposal for Vector API masking support. 

This is the masking part of https://bugs.openjdk.java.net/browse/JDK-8261663 - the second incubator for JEP 338 Vector API. As the JEP described, the masked vector operations are currently implemented by composing the non-masked operation with a blend operation. This can be improved by using the hardware mask feature on supported architecture like Arm SVE and X86 AVX-512. So here is the proposals for Arm SVE.  We assume the ideas could also be applied to X86 AVX-512.

To support the masking feature, this PR added the following implementations:
  -  SVE predicate register allocation
  -  Mask type and basic mask IR definition
  -  Mask implementation for masked vector store
  -  Mask implementation for masked binary operations

For the masked binary operations, we have created two proposals for discussion:
  -  By mainly changing the C2 compiler
  -  By improving the Vector API Java implementation together with simpler C2 compiler changes
  
This PR shows the second solution since we think this solution is better.  But we also have a prototype for the first solution. Please see: https://github.com/XiaohongGong/panama-vector/commit/372489feeae06bc53c46709d389cb0e46e9fb4f6 . The basic support changes are shared with this PR.

This PR doesn't contain all the masking support changes. There are still too many missing parts that we will continue working on, including:
  -  Mask support for other operations (unary,ternary,reduction,load,etc.)
  -  More mask IRs implementation (maskAll, toVector, allTrue, anyTrue, trueCount, eq, etc)
  -  Better solution for vector mask load/store  (the memory type is boolean)
  -  Vector boxing/unboxing support for mask type (deoptimization?)
  -  Tail loop elimination?

It's worth to mention that this PR mainly provides the proposals for SVE masking support, and any suggestions and discussions are welcome! Thanks a lot!

---------
### Progress

### Contributors
 * Ningsheng Jian `<njian@openjdk.org>`
 * Xiaohong Gong `<xgong@openjdk.org>`

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-vector pull/40/head:pull/40`
`$ git checkout pull/40`

To update a local copy of the PR:
`$ git checkout pull/40`
`$ git pull https://git.openjdk.java.net/panama-vector pull/40/head`

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64_sve.ad
+++ b/src/hotspot/cpu/aarch64/aarch64_sve.ad
@@ -220,17 +220,6 @@ source %{
     }
   }
 
-  static void sve_compare(C2_MacroAssembler masm, PRegister pd, BasicType bt,
-                          PRegister pg, FloatRegister zn, FloatRegister zm, int cond) {
-    Assembler::SIMD_RegVariant size = elemType_to_regVariant(bt);
-    if (bt == T_FLOAT || bt == T_DOUBLE) {
-      sve_compare_fp(masm, pd, size, pg, zn, zm, cond);
-    } else {
-      assert(is_integral_type(bt), "Unsupported type");
-      sve_compare_integral(masm, pd, size, pg, zn, zm, cond);
-    }
-  }
-
   bool op_sve_supported(int opcode) {
     switch (opcode) {
       case Op_MulAddVS2VI:
@@ -386,11 +375,11 @@ instruct vector2vmask(pRegGov pg, vReg src, rFlagsReg cr) %{
   match(Set pg (VectorToMask src));
   effect(KILL cr, DEF pg, USE src);
   ins_cost(SVE_COST);
-  format %{ "sve_cmpeq  $pg, $src, -1\t# vector mask (sve)" %}
+  format %{ "sve_cmpne  $pg, $src, 0\t# vector mask (sve)" %}
   ins_encode %{
     BasicType bt = vector_element_basic_type(this, $src);
-    __ sve_cmpeq(as_PRegister($pg$$reg), elemType_to_regVariant(bt),
-                 ptrue, as_FloatRegister($src$$reg), -1);
+    __ sve_cmpne(as_PRegister($pg$$reg), elemType_to_regVariant(bt),
+                 ptrue, as_FloatRegister($src$$reg), 0);
   %}
   ins_pipe(pipe_slow);
 %}
@@ -456,11 +445,11 @@ instruct vmask2vector(vReg dst, pRegGov pg) %{
   predicate(UseSVE > 0);
   match(Set dst (MaskToVector pg));
   ins_cost(SVE_COST);
-  format %{ "sve_cpy $dst, $pg, -1\t# mask to vector (sve)" %}
+  format %{ "sve_cpy $dst, $pg, 1\t# mask to vector (sve)" %}
   ins_encode %{
     BasicType bt = vector_element_basic_type(this);
     __ sve_cpy(as_FloatRegister($dst$$reg), elemType_to_regVariant(bt),
-               as_PRegister($pg$$reg), -1, false);
+               as_PRegister($pg$$reg), 1, false);
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1538,26 +1527,6 @@ instruct vpopcountI(vReg dst, vReg src) %{
   ins_pipe(pipe_slow);
 %}
 
-// vector mask compare
-
-instruct vmaskcmp(vReg dst, vReg src1, vReg src2, immI cond, pRegGov pTmp, rFlagsReg cr) %{
-  predicate(UseSVE > 0 && n->as_Vector()->length_in_bytes() >= 16);
-  match(Set dst (VectorMaskCmp (Binary src1 src2) cond));
-  effect(TEMP pTmp, KILL cr);
-  ins_cost(2 * SVE_COST);
-  format %{ "sve_cmp $pTmp, $src1, $src2\n\t"
-            "sve_cpy $dst, $pTmp, -1\t # vector mask cmp (sve)" %}
-  ins_encode %{
-    BasicType bt = vector_element_basic_type(this);
-    sve_compare(C2_MacroAssembler(&cbuf), as_PRegister($pTmp$$reg), bt,
-                ptrue, as_FloatRegister($src1$$reg),
-                as_FloatRegister($src2$$reg), (int)$cond$$constant);
-    __ sve_cpy(as_FloatRegister($dst$$reg), elemType_to_regVariant(bt),
-               as_PRegister($pTmp$$reg), -1, false);
-  %}
-  ins_pipe(pipe_slow);
-%}
-
 // vector blend
 
 instruct vblend(vReg dst, vReg src1, vReg src2, pRegGov pg) %{
@@ -1580,11 +1549,10 @@ instruct vloadmaskB(vReg dst, vReg src) %{
   predicate(UseSVE > 0 && n->as_Vector()->length_in_bytes() >= 16 &&
             n->bottom_type()->is_vect()->element_basic_type() == T_BYTE);
   match(Set dst (VectorLoadMask src));
-  ins_cost(SVE_COST);
-  format %{ "sve_neg $dst, $src\t # vector load mask (B)" %}
+  ins_cost(0);
+  format %{ "vloadmaskB (elided)\t# vector load mask (sve) (B) - do nothing" %}
   ins_encode %{
-    __ sve_neg(as_FloatRegister($dst$$reg), __ B, ptrue,
-               as_FloatRegister($src$$reg));
+    // empty
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1593,14 +1561,11 @@ instruct vloadmaskS(vReg dst, vReg src) %{
   predicate(UseSVE > 0 && n->as_Vector()->length_in_bytes() >= 16 &&
             n->bottom_type()->is_vect()->element_basic_type() == T_SHORT);
   match(Set dst (VectorLoadMask src));
-  ins_cost(2 * SVE_COST);
-  format %{ "sve_uunpklo $dst, $src\n\t"
-            "sve_neg $dst, $dst\t # vector load mask (B to H)" %}
+  ins_cost(SVE_COST);
+  format %{ "sve_uunpklo $dst, $src\t# vector load mask (sve) (B to H)" %}
   ins_encode %{
     __ sve_uunpklo(as_FloatRegister($dst$$reg), __ H,
                    as_FloatRegister($src$$reg));
-    __ sve_neg(as_FloatRegister($dst$$reg), __ H, ptrue,
-               as_FloatRegister($dst$$reg));
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1610,17 +1575,14 @@ instruct vloadmaskI(vReg dst, vReg src) %{
             (n->bottom_type()->is_vect()->element_basic_type() == T_INT ||
              n->bottom_type()->is_vect()->element_basic_type() == T_FLOAT));
   match(Set dst (VectorLoadMask src));
-  ins_cost(3 * SVE_COST);
+  ins_cost(2 * SVE_COST);
   format %{ "sve_uunpklo $dst, $src\n\t"
-            "sve_uunpklo $dst, $dst\n\t"
-            "sve_neg $dst, $dst\t # vector load mask (B to S)" %}
+            "sve_uunpklo $dst, $dst\t# vector load mask (sve) (B to S)" %}
   ins_encode %{
     __ sve_uunpklo(as_FloatRegister($dst$$reg), __ H,
                    as_FloatRegister($src$$reg));
     __ sve_uunpklo(as_FloatRegister($dst$$reg), __ S,
                    as_FloatRegister($dst$$reg));
-    __ sve_neg(as_FloatRegister($dst$$reg), __ S, ptrue,
-               as_FloatRegister($dst$$reg));
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1630,11 +1592,10 @@ instruct vloadmaskL(vReg dst, vReg src) %{
             (n->bottom_type()->is_vect()->element_basic_type() == T_LONG ||
              n->bottom_type()->is_vect()->element_basic_type() == T_DOUBLE));
   match(Set dst (VectorLoadMask src));
-  ins_cost(4 * SVE_COST);
+  ins_cost(3 * SVE_COST);
   format %{ "sve_uunpklo $dst, $src\n\t"
             "sve_uunpklo $dst, $dst\n\t"
-            "sve_uunpklo $dst, $dst\n\t"
-            "sve_neg $dst, $dst\t # vector load mask (B to D)" %}
+            "sve_uunpklo $dst, $dst\t# vector load mask (sve) (B to D)" %}
   ins_encode %{
     __ sve_uunpklo(as_FloatRegister($dst$$reg), __ H,
                    as_FloatRegister($src$$reg));
@@ -1642,8 +1603,6 @@ instruct vloadmaskL(vReg dst, vReg src) %{
                    as_FloatRegister($dst$$reg));
     __ sve_uunpklo(as_FloatRegister($dst$$reg), __ D,
                    as_FloatRegister($dst$$reg));
-    __ sve_neg(as_FloatRegister($dst$$reg), __ D, ptrue,
-               as_FloatRegister($dst$$reg));
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1653,11 +1612,10 @@ instruct vloadmaskL(vReg dst, vReg src) %{
 instruct vstoremaskB(vReg dst, vReg src, immI_1 size) %{
   predicate(UseSVE > 0 && n->as_Vector()->length() >= 16);
   match(Set dst (VectorStoreMask src size));
-  ins_cost(SVE_COST);
-  format %{ "sve_neg $dst, $src\t # vector store mask (B)" %}
+  ins_cost(0);
+  format %{ "vstoremaskB (elided)\t# vector store mask (sve) (B) - do nothing" %}
   ins_encode %{
-    __ sve_neg(as_FloatRegister($dst$$reg), __ B, ptrue,
-               as_FloatRegister($src$$reg));
+    // empty
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1666,17 +1624,13 @@ instruct vstoremaskS(vReg dst, vReg src, vReg tmp, immI_2 size) %{
   predicate(UseSVE > 0 && n->as_Vector()->length() >= 8);
   match(Set dst (VectorStoreMask src size));
   effect(TEMP_DEF dst, TEMP tmp);
-  ins_cost(3 * SVE_COST);
+  ins_cost(2 * SVE_COST);
   format %{ "sve_dup $tmp, 0\n\t"
-            "sve_uzp1 $dst, $src, $tmp\n\t"
-            "sve_neg $dst, $dst\t # vector store mask (sve) (H to B)" %}
+            "sve_uzp1 $dst, $src, $tmp\t# vector store mask (sve) (H to B)" %}
   ins_encode %{
     __ sve_dup(as_FloatRegister($tmp$$reg), __ H, 0);
     __ sve_uzp1(as_FloatRegister($dst$$reg), __ B,
                 as_FloatRegister($src$$reg), as_FloatRegister($tmp$$reg));
-    __ sve_neg(as_FloatRegister($dst$$reg), __ B, ptrue,
-               as_FloatRegister($dst$$reg));
-
   %}
   ins_pipe(pipe_slow);
 %}
@@ -1685,19 +1639,16 @@ instruct vstoremaskI(vReg dst, vReg src, vReg tmp, immI_4 size) %{
   predicate(UseSVE > 0 && n->as_Vector()->length() >= 4);
   match(Set dst (VectorStoreMask src size));
   effect(TEMP_DEF dst, TEMP tmp);
-  ins_cost(4 * SVE_COST);
+  ins_cost(3 * SVE_COST);
   format %{ "sve_dup $tmp, 0\n\t"
             "sve_uzp1 $dst, $src, $tmp\n\t"
-            "sve_uzp1 $dst, $dst, $tmp\n\t"
-            "sve_neg $dst, $dst\t # vector store mask (sve) (S to B)" %}
+            "sve_uzp1 $dst, $dst, $tmp\t# vector store mask (sve) (S to B)" %}
   ins_encode %{
     __ sve_dup(as_FloatRegister($tmp$$reg), __ S, 0);
     __ sve_uzp1(as_FloatRegister($dst$$reg), __ H,
                 as_FloatRegister($src$$reg), as_FloatRegister($tmp$$reg));
     __ sve_uzp1(as_FloatRegister($dst$$reg), __ B,
                 as_FloatRegister($dst$$reg), as_FloatRegister($tmp$$reg));
-    __ sve_neg(as_FloatRegister($dst$$reg), __ B, ptrue,
-               as_FloatRegister($dst$$reg));
   %}
   ins_pipe(pipe_slow);
 %}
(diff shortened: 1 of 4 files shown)


XiaohongGong:
Documents:
http://cr.openjdk.java.net/~xgong/rfr/mask/Vector%20API%20masking%20support%20proposal%20for%20Arm%20SVE.pdf
http://cr.openjdk.java.net/~xgong/rfr/mask/VectorAPI%20masking%20support.pdf

XiaohongGong:
Hi there, could anyone please take a look at this PR and the proposals? Any feedback is welcome! Thanks a lot!

DamonFool:
Good job! 
Thanks @XiaohongGong  and @nsjian .

Just wondering: is there any performance number before and after this enhancement?

There seems some code duplication between inline_vector_nary_operation and inline_vector_nary_mask_operation.
Is it possible to support mask operations in inline_vector_nary_operation(int n)?

Thanks.

XiaohongGong:
> Good job!
> Thanks @XiaohongGong and @nsjian .
> 
> Just wondering: is there any performance number before and after this enhancement?

Thanks for looking at this PR @DamonFool . Currently we only runs a small part of JMH benchmarks (`ADDMasked`) internally. And since the existing cases are simple which focus on single API operations, the performance can improve but the number is not so obvious (about 6% ~ 13% for integer types). Also we still have many missing masking supports, so we think it's too early to run all the benchmarks. 

> There seems some code duplication between inline_vector_nary_operation and inline_vector_nary_mask_operation.
> Is it possible to support mask operations in inline_vector_nary_operation(int n)?

This is a temporary state since we only tries the masking binary with this solutions. So yes, these two functions can be merged into one. And we will do this once unary/ternary masked operations are supported in future. Thanks!

> 
> Thanks.

DamonFool:
> And since the existing cases are simple which focus on single API operations, the performance can improve but the number is not so obvious (about 6% ~ 13% for integer types). Also we still have many missing masking supports, so we think it's too early to run all the benchmarks.

Good to know that.
Thanks.

XiaohongGong:
Hi @sviswa7 @jatin-bhateja ,

Could you please take a look at this PR which contains the basic masking support? Although the changes focus on Arm SVE platform, we expect the ideas can also be applied to AVX-512 especially for the shared codes. It will be very helpful if you can give any comments from the side of AVX-512. And it will be very great if there is an optimal final solution for the masking support both for Arm SVE and AVX-512. So please help to have a look at it! Thanks so much! 

BTW, this PR only contains a part of the masking support codes. We will continue working on the missing parts including other masked vector operations, mask generation operations and the mask operations. And currently we are looking at the deoptimization support for predicate registers. To avoid conflicts, it will be helpful if you could also share us any ideas or plans from your side. Thanks so much!

Best Regards,
Xiaohong Gong

sviswa7:
Hi Xiaohong (@XiaohongGong) ,
Jatin is working on masking support. He will share his feedback, ideas and plans shortly with you.
Best Regards,
Sandhya

XiaohongGong:
> Hi Xiaohong (@XiaohongGong) ,
> Jatin is working on masking support. He will share his feedback, ideas and plans shortly with you.
> Best Regards,
> Sandhya

Hi @sviswa7 , thanks so much for telling this. It's really great if we can persistently share with each other.

XiaohongGong:
Hi, I have to force push all the codes again to resolve the conflicts due to the latest merging from jdk master. Thanks!

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -339,6 +339,52 @@ instruct storeV_partial(vReg src, vmemA mem, pRegGov pTmp, iRegINoSp tmp, rFlags
...
+                as_FloatRegister($src2$$reg), (int)$cond$$constant);
+  %}
+  ins_pipe(pipe_slow);
+%}
+
+// mask to vector
+
+instruct vmask2vector(vReg dst, pRegGov pg) %{
+  predicate(UseSVE > 0);
+  match(Set dst (MaskToVector pg));
```
jatin-bhateja, line 438: We already have VectorLoadMask and VectorStoreMask nodes which have same semantics as Mask2Vector and Vector2Mask operation. Can you please elaborate the need for these new nodes.

src/hotspot/share/opto/type.cpp
```diff
@@ -679,13 +680,17 @@ void Type::Initialize_shared(Compile* current) {
     TypeVect::VECTZ = TypeVect::make(T_FLOAT,16);
   }
 
+  if (Matcher::has_predicated_vectors()) {
+    TypeVMask::VMASK = TypeVMask::make(T_BYTE, Matcher::max_vector_size(T_BYTE));
+  }
+
```
jatin-bhateja, line 686: Instead of creating a new type altogether,  we can extended the mask type from the existing TypeVect class.
This is being done for X86 base patch, currently under review:
https://github.com/openjdk/jdk/pull/2768

I think same type (TypeVectMask) can be used for SVE support.

src/hotspot/share/opto/type.hpp
```diff
@@ -845,6 +849,39 @@ class TypeVectZ : public TypeVect {
...
+  }
+  uint length(void) const {
+    return _length;
+  }
+  virtual bool empty(void) const;
+  virtual bool eq(const Type* t) const;
+  virtual int hash() const;  // Type specific hashing
+  virtual bool singleton(void) const;
+  virtual const Type* xdual() const;
+  virtual const Type* xmeet(const Type* t) const;
```
jatin-bhateja, line 882: Extending TypeVector for creating new mask type seems like a logical extension, this has been done for AVX-512 predicated register support, it could be used for SVE also once the patch lands into mainline.

https://github.com/openjdk/jdk/pull/2768

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -854,6 +862,24 @@ class VectorToMaskNode : public VectorMaskNode {
   virtual Node* Identity(PhaseGVN* phase);
 };
 
+// Vector compare node with a TypeVMask bottom_type. It is specially generated for platforms
+// that have mask hardware feature. The main difference with "VectorMaskCmpNode" is that this
+// is a kind of mask node, while "VectorMaskCmpNode" is a vector node.
+class VectorCmpMaskGenNode : public VectorMaskNode {
+ public:
+  VectorCmpMaskGenNode(Node* in1, Node* in2, ConINode* predicate_node, const TypeVMask* vmask_type) :
```
jatin-bhateja, line 918: VectorMaskCmpNode compares two vectors and produces comparison results into a vector (for targets which does not supports predicated vectors) or a predicate vector/register.  Can you kindly elaborate the need for this new node.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
```diff
@@ -971,40 +985,36 @@ instruct vstoremaskL(vReg dst, vReg src, vReg tmp, immI_8 size) %{
...
-instruct vloadmask_loadV(vReg dst, vmemA mem) %{
-  predicate(UseSVE > 0 && n->as_Vector()->length_in_bytes() >= 16 &&
-            n->in(1)->bottom_type()->is_vect()->element_basic_type() == T_BOOLEAN);
-  match(Set dst (VectorLoadMask (LoadVector mem)));
-  ins_cost(2 * SVE_COST);
-  format %{ "sve_ld1b $dst, $mem\n\t"
-            "sve_neg $dst, $dst\t # load vector mask (sve)" %}
+instruct loadVMask(vReg dst, vmemA mem) %{
+  predicate(UseSVE > 0);
+  match(Set dst (LoadVectorMask mem));
```
jatin-bhateja, line 1121: Are we trying to perform a vector load operation under the influence of a mask here? 
C2 already has an ideal node LoadVectorMaskedNode which may come handy.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
```diff
@@ -971,40 +985,36 @@ instruct vstoremaskL(vReg dst, vReg src, vReg tmp, immI_8 size) %{
...
-  predicate(UseSVE > 0 && n->as_StoreVector()->length() >= 2 &&
-            n->as_StoreVector()->vect_type()->element_basic_type() == T_BOOLEAN);
-  match(Set mem (StoreVector mem (VectorStoreMask src size)));
-  effect(TEMP tmp);
-  ins_cost(2 * SVE_COST);
-  format %{ "sve_neg $tmp, $src\n\t"
-            "sve_st1b $tmp, $mem\t # store vector mask (sve)" %}
+instruct storeVMask(vReg src, vmemA mem) %{
+  predicate(UseSVE > 0);
+  match(Set mem (StoreVectorMask mem src));
```
jatin-bhateja, line 1139: Similar comment as above, existing C2 ideal node StoreVectorMaskedNode could be used here.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -1277,40 +1291,36 @@ instruct vstoremaskL(vReg dst, vReg src, vReg tmp, immI_8 size) %{
...
-  format %{ "sve_ld1b $dst, $mem\n\t"
-            "sve_neg $dst, $dst\t # load vector mask (sve)" %}
+instruct loadVMask(vReg dst, vmemA mem) %{
+  predicate(UseSVE > 0);
+  match(Set dst (LoadVectorMask mem));
+  ins_cost(SVE_COST);
+  format %{ "sve_ld1b $dst, $mem\t# load vector mask (sve)" %}
   ins_encode %{
+    // Load mask values which are boolean type, and extend them to the
+    // expected vector element type.
```
jatin-bhateja, line 1719: Isn't this same as existing VectorLoadMaskNode ?

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -1277,40 +1291,36 @@ instruct vstoremaskL(vReg dst, vReg src, vReg tmp, immI_8 size) %{
...
-            n->as_StoreVector()->vect_type()->element_basic_type() == T_BOOLEAN);
-  match(Set mem (StoreVector mem (VectorStoreMask src size)));
-  effect(TEMP tmp);
-  ins_cost(2 * SVE_COST);
-  format %{ "sve_neg $tmp, $src\n\t"
-            "sve_st1b $tmp, $mem\t # store vector mask (sve)" %}
+instruct storeVMask(vReg src, vmemA mem) %{
+  predicate(UseSVE > 0);
+  match(Set mem (StoreVectorMask mem src));
+  format %{ "sve_st1b $src, $mem\t # store vector mask (sve)" %}
```
jatin-bhateja, line 1733: Same comments as above, how different is this from existing VectorStoreMaskNode

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -2467,6 +2468,18 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType
   return true; // Per default match rules are supported.
 }
 
+const bool Matcher::match_rule_supported_masked_vector(int opcode, int vlen, BasicType bt) {
+  if (!match_rule_supported(opcode) || !vector_size_supported(bt, vlen)) {
```
jatin-bhateja, line 2472: Can we not just use existing routine match_rule_supported_vector here?

src/hotspot/cpu/aarch64/aarch64_sve.ad
XiaohongGong: Thanks for your comment about this. `VectorLoadMask` and `VectorStoreMask` are vector nodes that have the `TypeVect` type, while `Vector2Mask` is a mask node that have the `TypeVMask` type.  For SVE, the two new added nodes are used for conversions between vector register and predicate register, e.g. `Mask2Vector` for `predicate -> vector` and `Vector2Mask` for `vector -> predicate`. Yes, it seems `VectorLoadMask` and `VectorStoreMask` can match the same requirements like what have been done in the AVX-512 patch. However, in our whole proposal, we defined different IRs to separate vector and mask operations. I think the main conflict is the IR part for mask. And it's better to get a final solution to the different operations together for SVE and AVX-512.

XiaohongGong: Note that `MaskToVector` can be used for API `VectorMask.toVector()` in future which is implemented with `blend` currently.

src/hotspot/share/opto/vectornode.hpp
XiaohongGong: `VectorCmpMaskGen` is a kind of mask IR which generates predicate register.  The same with `VectorToMask`, we defined a series of mask IRs that are different with the original vector IRs.

src/hotspot/cpu/aarch64/aarch64_sve.ad
XiaohongGong: I'm not sure whether this can benefit AVX-512 masking load. It loads the values from boolean memory and extends the boolean values to relative other basic types (short,int,long). Different with AVX-512, SVE predicate register controls the vector activity depending on the element type, i.e. it uses 1-bit to control a byte lane, and 2-bits for a short lane, and 4-bits for an int lane, and so on. The type extending/narrowing between real basic type and boolean is always needed. As SVE provides a single instruction to implement the `load+extend` function, we don't need to load values and extend them with external instructions. So that's what the new added node used to do. Maybe a better name for it is `LoadVectorExtendNode`.

src/hotspot/cpu/aarch64/aarch64_sve.ad
XiaohongGong: The same with `LoadVectorMaskNode`.

src/hotspot/cpu/aarch64/aarch64.ad
XiaohongGong: This can be improved in future. But I think it's different with `match_rule_supported_vector`. It's used to check whether the optimized predicate version is supported/implemented for an operation.

XiaohongGong: I guess there might be different masking support status for operations with SVE and AVX-512. So the check that depends on the architecture should be implemented in the backend.

src/hotspot/share/opto/type.cpp
XiaohongGong: Yes, there is some discussions focus on the mask type definition in jdk/jdk master. It's a good idea to define a subclass of `TypeVect`. It seems this can be used for SVE, but I'm not quite sure and we will have a try internally. Hope we can get a final point on the mask type first. Thanks!

jatin-bhateja:
Hi @XiaohongGong , @nsjian 

I have query regarding scalable vectors operand(vReg):  Ideally type of machine node must comply with the type of its def machine operand,  for scalable vector vReg::type() will always return a TypeVect::VECA (maximal vector size supported) irrespective of the type of MachTypeNode, this works fine because bottom_type() for MachTypeNode returns the type propagated to it from Ideal node during selection.
  
There are two places where type this in-consistency may show up:-
1) If bottom type of machprojnode of a mask generating node is queried. 
2) If someone directly queries the bottom type of def operand (ideally this should be rare)

We can fix 1) by returning the bottom type of incoming node of a MachProjNode. Do you foresee any penalty due to it, I am not able to think of one. 

Second observation (though not related to X86 or Masked operations),  it seems you are spilling complete scalable vector (2048 bits) (MachSpillCopyNode::implementation)? 
I think resolving generic scalable vectors (2048 bits) to concrete ones (tightly coupled to machnode type) may prevent spilling extra bits. 

Best Regards,
Jatin

nsjian:
Hi @jatin-bhateja 

> I have query regarding scalable vectors operand(vReg): Ideally type of machine node must comply with the type of its def machine operand, for scalable vector vReg::type() will always return a TypeVect::VECA (maximal vector size supported) irrespective of the type of MachTypeNode, this works fine because bottom_type() for MachTypeNode returns the type propagated to it from Ideal node during selection.
> 
> There are two places where this type in-consistency may show up:-
> 
> 1. If bottom type of machprojnode of a mask generating node is queried.
> 2. If someone directly queries the bottom type of def operand (ideally this should be rare)
> 
> We can fix 1) by returning the bottom type of incoming node of a MachProjNode. Do you foresee any penalty due to current inconsistencies, I am not able to think of one.
> 

I don't quite understand this inconsistency. Could you please elaborate a bit more? I am expecting a mask generating node would always get a TypeVMask (TypeVectMask) on supporting platform, which is pRegGovOper on SVE.

> Second observation (though not related to X86 or Masked operations), it seems you are spilling complete scalable vector (2048 bits) (MachSpillCopyNode::implementation)?

We are spilling the whole SVE register, which could be 128 bits to 2048 bits, depending on current running hardware implementation. So, the spilling size could be different depending on current SVE vector size of running process, while not the architecturally max (2048) bits. 

> I think resolving generic scalable vectors (2048 bits) to concrete ones (tightly coupled to machnode type) may prevent spilling extra bits.
> 

Thanks for looking at our backend implementation details and the suggestion. The generic vector register implementation sounds a good idea. However, for SVE, architecturally we have only one concrete vector type, SVE scalable vector register, though its size depends on different hardware implementations.

Thanks,
Ningsheng

src/hotspot/cpu/aarch64/aarch64_sve.ad
jatin-bhateja: > Thanks for your comment about this. `VectorLoadMask` and `VectorStoreMask` are vector nodes that have the `TypeVect` type, while `Vector2Mask` is a mask node that have the `TypeVMask` type. For SVE, the two new added nodes are used for conversions between vector register and predicate register, e.g. `Mask2Vector` for `predicate -> vector` and `Vector2Mask` for `vector -> predicate`. Yes, it seems `VectorLoadMask` and `VectorStoreMask` can match the same requirements like what have been done in the AVX-512 patch. However, in our whole proposal, we defined different IRs to separate vector and mask operations. I think the main conflict is the IR part for mask. And it's better to get a final solution to the different operations together for SVE and AVX-512.

Hi @XiaohongGong , 
Yes currently all the ideal nodes have fixed types , but technically there is nothing incorrect in having nodes (VectorLoadMask/VectorStoreMask) with types based on the target features, e.g. VectorLoadMask may have TypeVect for targets which do-not support predicate registers and a predicate vector type (TypeVectMask/TypeVMask) otherwise. 
By taking this approach we may be able to save this synthetic duplication for masked flavors of vector operation nodes.

jatin-bhateja:
> I don't quite understand this inconsistency. Could you please elaborate a bit more? I am expecting a mask generating node would always get a TypeVMask (TypeVectMask) on supporting platform, which is pRegGovOper on SVE.

Hi @nsjian ,
This question was related to SVE support,
Function vRegOper::type() always returns Type::VECA.
Type::VECA is initialized as maximal sized Scalable vector i.e. its size should be 2048 bit.  For a MachTypeNode having vReg as its def operand, will the def operand's type always match the bottom_type() of MachTypeNode?

nsjian:
Hi @jatin-bhateja ,

> Type::VECA is initialized as maximal sized Scalable vector i.e. its size should be 2048 bit. For a MachTypeNode having vReg as its def operand, will the def operand's type always match the bottom_type() of MachTypeNode?

Yes, I think so. On SVE target, we have only one vector type, TypeVect::VECTA and no other vector types. And it's real size is runtime detected, based on hardware implementation.

Thanks,
Ningsheng

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -376,6 +376,227 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
...
+  }
+
+  // Currently just focus on binary mask operation.
+  if (n == 2 && mask != NULL) {
+    if (sopc != 0 &&
+        Matcher::match_rule_supported_masked_vector(sopc, num_elem, elem_bt) &&
+        Matcher::match_rule_supported(Op_VectorToMask)) {
+      const TypeVMask* vmask_type = TypeVMask::make(elem_bt, num_elem);
+      mask = gvn().transform(new VectorToMaskNode(mask, vmask_type));
+      operation->add_req(mask);
```
jatin-bhateja, line 583: We are adding a new input to a binary operation with existing opcodes i.e.
Eg. following will be graph shapes with or w/o mask node.
   AddVI SRC1 SRC2
   AddVI SRC1 SRC2 MASK

We may be able to curtail the number of instruction patterns for masked operation by folding blend + vector operation into a new VectorMaskedOperNode.  Effectively this way we may be able to remove lots of redundant instruction patterns from the AD files, which has been a cause of concern for us(X86) in past.

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -266,6 +266,20 @@ VM binaryOp(int oprId, Class<? extends VM> vmClass, Class<?> elementType, int le
         return defaultImpl.apply(vm1, vm2);
     }
 
+    @IntrinsicCandidate
+    public static
+    <V, M>
+    V binaryMaskOp(int oprId, Class<? extends V> vmClass, Class<? extends M> maskClass,
```
jatin-bhateja, line 272: Current patch make VectorSupport.binaryMaskOp the fundamental intrinsic for creating  a masked/non-masked vector nodes. Can you kindly elaborate the need for these change.

I think an ideal transformation in compiler should be able to save any changes proposed in java code, unless there is any pressing need for it.

src/hotspot/share/opto/vectorIntrinsics.cpp
jatin-bhateja, line 583: Following is the link to a reference implementation (PoC) for X86: 
http://cr.openjdk.java.net/~jbhateja/avx512_masked_operation_optimization/webrev.02/src/hotspot/share/opto/vectornode.cpp.udiff.html
This demonstrates creation of new masked operation node which folds blend+vector operation graph patterns, there is one new Ideal node for each kind of operation i.e. ternary, binary and unary. New node also carry the meta data info which specifies the kind of masked operation being performed.

iwanowww: One downside of `AddVI SRC1 SRC2 MASK` shape is that the operation is not commutative anymore:
```
AddVI SRC1 SRC2 MASK = VectorBlend SRC1 (AddVI SRC1 SRC2) MASK
AddVI SRC2 SRC1 MASK = VectorBlend SRC2 (AddVI SRC2 SRC1) MASK
```

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
XiaohongGong: We proposed two solutions for generating the masked vector operations. Changing the intrinsic is one of them. Another is the same with your idea (please see the prototype: https://github.com/XiaohongGong/panama-vector/commit/372489feeae06bc53c46709d389cb0e46e9fb4f6). Actually we think both solutions can work. But they all have their advantages and shortages.

src/hotspot/share/opto/vectorIntrinsics.cpp
XiaohongGong: Thanks for looking at this PT @iwanowww ! Yes, I agree that this can change the commutative feature by appending the mask. I found a potential influence during matching. The codes look like:
```
  void MatchNode::count_commutative_op(int& count) {
  static const char *commut_op_list[] = {
    "AddI","AddL","AddF","AddD",
    "AddVB","AddVS","AddVI","AddVL","AddVF","AddVD",
    "AndI","AndL",
    "AndV",
    "MaxI","MinI","MaxF","MinF","MaxD","MinD",
    "MaxV", "MinV",
    "MulI","MulL","MulF","MulD",
    "MulVB","MulVS","MulVI","MulVL","MulVF","MulVD",
    "MinV","MaxV",
    "OrI","OrL",
    "OrV",
    "XorI","XorL",
    "XorV"
  };
  int cnt = sizeof(commut_op_list)/sizeof(char*);

  if( _lChild && _rChild && (_lChild->_lChild || _rChild->_lChild) ) {
    // Don't swap if right operand is an immediate constant.
    bool is_const = false;
    if( _rChild->_lChild == NULL && _rChild->_rChild == NULL ) {
      FormDict &globals = _AD.globalNames();
      const Form *form = globals[_rChild->_opType];
      if ( form ) {
        OperandForm  *oper = form->is_operand();
        if( oper && oper->interface_type(globals) == Form::constant_interface )
          is_const = true;
      }
    }
    if( !is_const ) {
      for( int i=0; i<cnt; i++ ) {
        if( strcmp(_opType, commut_op_list[i]) == 0 ) {
          count++;
          _commutative_id = count; // id should be > 0
          break;
        }
      }
    }
  }
  if( _lChild )
    _lChild->count_commutative_op(count);
  if( _rChild )
    _rChild->count_commutative_op(count);
 }
```
For such issues, is it possible to add the `commutative` feature to a matchnode, or adding more conditions when checking whether the node is commutative? Regarding to the masked vector representation, do you have any better idea? We also tried with the `VectorMaskWrapperNode` before, but it's not better with some reasons that I pointed out at page-5 in http://cr.openjdk.java.net/~xgong/rfr/mask/Vector%20API%20masking%20support%20proposal%20for%20Arm%20SVE.pdf.  Thanks!

src/hotspot/cpu/aarch64/aarch64_sve.ad
XiaohongGong: This new node is performance friendly to SVE. For SVE, we changed the original `VectorLoadMask (LoadVector mem)` to `VectorToMask (LoadVectorMask mem)`. And the same logic to mask store.

XiaohongGong:
Hi, I added the predicate support for API `maskAll` and also removed the redundant `negations`. Thanks!

XiaohongGong:
I'v opened another PR for the masking support. See https://github.com/openjdk/panama-vector/pull/57

PR #41, opened 2021-02-27
8262492: Add cast nodes between float types implementation for Arm SVE

Wanghuang-Huawei:
Add cast nodes like castf2d castd2f

* add `sve_fcvt`
* reuse macro `VECTOR_CAST_I2F_NARROW1` `VECTOR_CAST_I2F_EXTEND1` (Should we change these macro names?)

other issues :
* precision loss

---------
### Progress

### Issue
 * [JDK-8262492](https://bugs.openjdk.java.net/browse/JDK-8262492): Add cast nodes between float types implementation for Arm SVE

### Reviewers

### Contributors
 * He Xuejin `<hexuejin2@huawei.com>`

### Download
`$ git fetch https://git.openjdk.java.net/panama-vector pull/41/head:pull/41`
`$ git checkout pull/41`

### DIFF
--- a/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
@@ -3356,7 +3356,7 @@ public:
 private:
 
   void encode_cvtf_T(SIMD_RegVariant T_dst, SIMD_RegVariant T_src,
-                     int& opc, int& opc2) {
+                     unsigned& opc, unsigned& opc2) {
     assert(T_src != B && T_dst != B &&
            T_src != Q && T_dst != Q, "invalid register variant");
     if (T_dst != D) {
@@ -3392,7 +3392,7 @@ public:
   void NAME(FloatRegister Zd, SIMD_RegVariant T_dst, PRegister Pg,      \
             FloatRegister Zn, SIMD_RegVariant T_src) {                  \
     starti;                                                             \
-    int opc, opc2;                                                      \
+    unsigned opc, opc2;                                                 \
     encode_cvtf_T(T_dst, T_src, opc, opc2);                             \
     f(0b01100101, 31, 24), f(opc, 23, 22), f(0b010, 21, 19);            \
     f(opc2, 18, 17), f(sign, 16), f(0b101, 15, 13);                     \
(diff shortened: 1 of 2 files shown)


src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3403,6 +3403,53 @@ void mvnw(Register Rd, Register Rm,
...
+        opc = 0b11;
+        opc2 = 0b00;
+      } else if (T_dst == S) {
+        opc = 0b11;
+        opc2 = 0b10;
+      }
+    }
+  }
+
+// SVE floating-point convert precision(predicated)
```
nsjian, line 3438: space between precision and '(' ?

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
```diff
@@ -3403,6 +3403,53 @@ void mvnw(Register Rd, Register Rm,
...
+            FloatRegister Zn, SIMD_RegVariant T_src) {                  \
+    starti;                                                             \
+    unsigned opc, opc2;                                                 \
+    encode_fcvt_T(T_src, T_dst, opc, opc2);                             \
+    f(0b01100101, 31, 24), f(opc, 23, 22), f(0b0010, 21, 18);           \
+    f(opc2, 17, 16), f(0b101, 15, 13);                                  \
+    pgrf(Pg, 10), rf(Zn, 5), rf(Zd, 0);                                 \
+  }
+
+  INSN(sve_fcvt);
```
nsjian, line 3450: Since there's only one use of the INSN macro, I think it's no need to define a macro here.

nsjian (commented):
> * reuse macro `VECTOR_CAST_I2F_NARROW1` `VECTOR_CAST_I2F_EXTEND1` (Should we change these macro names?)

I think it would be better to change the macro name.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: OK. I will change that.

src/hotspot/cpu/aarch64/assembler_aarch64.hpp
Wanghuang-Huawei: OK. I will clear that.

nsjian (approved):
Looks good to me.

PR #57, opened 2021-04-02
8264563: Add masked vector intrinsics for binary/store operations

XiaohongGong:
Hi, this is the basic masking support PR for Vector API mask operations on platforms like SVE/AVX-512. The main codes are from [1], which contains:

- The predicate register allocation for Arm SVE, and vector mask type definition.
- The basic optimization for parts of the mask operations with masking feature. It contains:   
 
  1. Vector API java implementation changes for masked binary/store.
  2. C2 compiler mid-end changes, including new vector intrinsics implementation and mask IRs.

Note that for easier discussion, this PR only provides the changes for limited masked operations (e.g. binary/store) and the mask generations (e.g. load/compare/maskAll). We will continue working on the following missing parts:

- Mask support for other operations (unary,ternary,reduction,load,etc.)
- More mask IRs implementation (and/or/xor, toVector, allTrue, anyTrue, trueCount, eq, etc)
- Vector boxing/unboxing support for mask type (deoptimization support for predicate registers)

Also note that this PR doesn't contain any backend implementations. So the blend pattern will be generated as before. Regarding to the AArch64 SVE backend support, we will create a separate PR based on this one in future.
  
[1] https://github.com/openjdk/panama-vector/pull/40

See more details from:
http://cr.openjdk.java.net/~xgong/rfr/mask/Vector%20API%20masking%20support%20proposal%20for%20Arm%20SVE.pdf
http://cr.openjdk.java.net/~xgong/rfr/mask/VectorAPI%20masking%20support.pdf

Any suggestions and discussions are welcome! Thanks a lot!

---------
### Progress

### Issue
 * [JDK-8264563](https://bugs.openjdk.java.net/browse/JDK-8264563): Add masked vector intrinsics for binary/store operations

### Reviewers
 * [Jatin Bhateja](https://openjdk.java.net/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to 9dd8856e26ef5d5eb7e8e2087011c9fde9b40b36
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to 9dd8856e26ef5d5eb7e8e2087011c9fde9b40b36
 * [Wang Huang](https://openjdk.java.net/census#whuang) (@Wanghuang-Huawei - Committer) ⚠️ Review applies to f7cd754dfd37bb9e1dd67cadd81f4e257b171aed

### Contributors
 * Ningsheng Jian `<njian@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/57/head:pull/57` \
`$ git checkout pull/57`

`$ git checkout pull/57` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/57/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 57`

`$ git pr show -t 57`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/57.diff">https://git.openjdk.java.net/panama-vector/pull/57.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -406,19 +406,18 @@ bool LibraryCallKit::inline_vector_nary_masked_operation(int n) {
   const TypeInstPtr* elem_klass   = gvn().type(argument(3))->isa_instptr();
   const TypeInt*     vlen         = gvn().type(argument(4))->isa_int();
 
-  if (opr == NULL || vector_klass == NULL || mask_klass == NULL || elem_klass == NULL || vlen == NULL ||
-      !opr->is_con() || vector_klass->const_oop() == NULL || mask_klass->const_oop() == NULL ||
-      elem_klass->const_oop() == NULL || !vlen->is_con()) {
+  if (opr == NULL || vector_klass == NULL || elem_klass == NULL || vlen == NULL ||
+      !opr->is_con() || vector_klass->const_oop() == NULL || elem_klass->const_oop() == NULL || !vlen->is_con()) {
     if (C->print_intrinsics()) {
-      tty->print_cr("  ** missing constant: opr=%s vclass=%s maskclass=%s etype=%s vlen=%s",
+      tty->print_cr("  ** missing constant: opr=%s vclass=%s etype=%s vlen=%s",
                     NodeClassNames[argument(0)->Opcode()],
                     NodeClassNames[argument(1)->Opcode()],
-                    NodeClassNames[argument(2)->Opcode()],
                     NodeClassNames[argument(3)->Opcode()],
                     NodeClassNames[argument(4)->Opcode()]);
     }
     return false; // not enough info for intrinsification
   }
+
   ciType* elem_type = elem_klass->const_oop()->as_instance()->java_mirror_type();
   if (!elem_type->is_primitive_type()) {
     if (C->print_intrinsics()) {
@@ -431,6 +430,33 @@ bool LibraryCallKit::inline_vector_nary_masked_operation(int n) {
       tty->print_cr("  ** klass argument not initialized");
     }
     return false;
+  }
+
+  // "argument(n + 5)" should be the mask object. We assume it is "null" when no mask
+  // is used to control this operation.
+  const Type* vmask_type = gvn().type(argument(n + 5));
+  bool is_masked_op = vmask_type != TypePtr::NULL_PTR;
+  if (is_masked_op) {
+    if (mask_klass == NULL || mask_klass->const_oop() == NULL) {
+      if (C->print_intrinsics()) {
+        tty->print_cr("  ** missing constant: maskclass=%s", NodeClassNames[argument(2)->Opcode()]);
+      }
+      return false; // not enough info for intrinsification
+    }
+
+    if (!is_klass_initialized(mask_klass)) {
+      if (C->print_intrinsics()) {
+        tty->print_cr("  ** mask klass argument not initialized");
+      }
+      return false;
+    }
+
+    if (vmask_type->maybe_null()) {
+      if (C->print_intrinsics()) {
+        tty->print_cr("  ** null mask values are not allowed for masked op");
+      }
+      return false;
+    }
   }
 
   BasicType elem_bt = elem_type->basic_type();
@@ -446,18 +472,8 @@ bool LibraryCallKit::inline_vector_nary_masked_operation(int n) {
   ciKlass* vbox_klass = vector_klass->const_oop()->as_instance()->java_lang_Class_klass();
   const TypeInstPtr* vbox_type = TypeInstPtr::make_exact(TypePtr::NotNull, vbox_klass);
 
-  // "argument(n + 5)" should be the mask object. We assume it is "null" when no mask
-  // is used to control this operation.
-  bool is_masked_op = gvn().type(argument(n + 5)) != TypePtr::NULL_PTR;
   if (is_vector_mask(vbox_klass)) {
     assert(!is_masked_op, "mask operations do not need mask to control");
-  }
-
-  if (is_masked_op && !is_klass_initialized(mask_klass)) {
-    if (C->print_intrinsics()) {
-      tty->print_cr("  ** mask klass argument not initialized");
-    }
-    return false;
   }
 
   if (opc == Op_CallLeafVector) {
@@ -548,7 +564,8 @@ bool LibraryCallKit::inline_vector_nary_masked_operation(int n) {
     }
 
     // Return true if current platform has implemented the masked operation with predicate feature.
-    use_predicate = sopc != 0 && Matcher::match_rule_supported_vector_masked(sopc, num_elem, elem_bt);
+    use_predicate = sopc != 0 && Matcher::has_predicated_vectors() &&
+                    Matcher::match_rule_supported_vector_masked(sopc, num_elem, elem_bt);
     if (!use_predicate && !arch_supports_vector(Op_VectorBlend, num_elem, elem_bt, VecMaskUseLoad)) {
       return false;
     }
@@ -582,7 +599,10 @@ bool LibraryCallKit::inline_vector_nary_masked_operation(int n) {
 
   if (is_masked_op && mask != NULL) {
     if (use_predicate) {
-      // TODO: add predicate implementation for masked operation.
+      if (C->print_intrinsics()) {
+        tty->print_cr("  ** predicate feature is not supported yet!");
+      }
+      return false;
     } else {
       operation = new VectorBlendNode(opd1, operation, mask);
     }
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
@@ -599,7 +599,6 @@ public abstract class ByteVector extends AbstractVector<Byte> {
     @ForceInline
     final
     ByteVector lanewiseTemplate(VectorOperators.Binary op,
-                                          Class<? extends VectorMask<Byte>> maskClass,
                                           Vector<Byte> v) {
         ByteVector that = (ByteVector) v;
         that.check(this);
@@ -636,7 +635,7 @@ public abstract class ByteVector extends AbstractVector<Byte> {
 
         int opc = opCode(op);
         return VectorSupport.binaryMaskedOp(
-            opc, getClass(), maskClass, byte.class, length(),
+            opc, getClass(), null, byte.class, length(),
             this, that, null,
             BIN_MASKED_IMPL.find(op, opc, ByteVector::binaryOperations));
     }
@@ -818,8 +817,7 @@ public abstract class ByteVector extends AbstractVector<Byte> {
         byte e1 = (byte) e;
         if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
-            ) {
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1);
@@ -842,8 +840,7 @@ public abstract class ByteVector extends AbstractVector<Byte> {
         byte e1 = (byte) e;
         if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
-            ) {
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
@@ -599,7 +599,6 @@ public abstract class IntVector extends AbstractVector<Integer> {
     @ForceInline
     final
     IntVector lanewiseTemplate(VectorOperators.Binary op,
-                                          Class<? extends VectorMask<Integer>> maskClass,
                                           Vector<Integer> v) {
         IntVector that = (IntVector) v;
         that.check(this);
@@ -636,7 +635,7 @@ public abstract class IntVector extends AbstractVector<Integer> {
 
         int opc = opCode(op);
         return VectorSupport.binaryMaskedOp(
-            opc, getClass(), maskClass, int.class, length(),
+            opc, getClass(), null, int.class, length(),
             this, that, null,
             BIN_MASKED_IMPL.find(op, opc, IntVector::binaryOperations));
     }
@@ -818,8 +817,7 @@ public abstract class IntVector extends AbstractVector<Integer> {
         int e1 = (int) e;
         if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
-            ) {
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1);
@@ -842,8 +840,7 @@ public abstract class IntVector extends AbstractVector<Integer> {
         int e1 = (int) e;
         if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
-            ) {
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
@@ -647,7 +647,6 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
     @ForceInline
     final
     $abstractvectortype$ lanewiseTemplate(VectorOperators.Binary op,
-                                          Class<? extends VectorMask<$Boxtype$>> maskClass,
                                           Vector<$Boxtype$> v) {
         $abstractvectortype$ that = ($abstractvectortype$) v;
         that.check(this);
@@ -694,7 +693,7 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
 
         int opc = opCode(op);
         return VectorSupport.binaryMaskedOp(
-            opc, getClass(), maskClass, $type$.class, length(),
+            opc, getClass(), null, $type$.class, length(),
             this, that, null,
             BIN_MASKED_IMPL.find(op, opc, $abstractvectortype$::binaryOperations));
     }
@@ -899,12 +898,13 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
     $abstractvectortype$ lanewise(VectorOperators.Binary op,
                                   long e) {
         $type$ e1 = ($type$) e;
-        if ((long)e1 != e
 #if[BITWISE]
+        if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
+#else[BITWISE]
+        if ((long)e1 != e) {
 #end[BITWISE]
-            ) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1);
@@ -925,12 +925,13 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
     $abstractvectortype$ lanewise(VectorOperators.Binary op,
                                   long e, VectorMask<$Boxtype$> m) {
         $type$ e1 = ($type$) e;
-        if ((long)e1 != e
 #if[BITWISE]
+        if ((long)e1 != e
             // allow shift ops to clip down their int parameters
-            && !(opKind(op, VO_SHIFT) && (int)e1 == e)
+            && !(opKind(op, VO_SHIFT) && (int)e1 == e)) {
+#else[BITWISE]
+        if ((long)e1 != e) {
 #end[BITWISE]
-            ) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
@@ -276,7 +276,7 @@ final class Byte128Vector extends ByteVector {
     @Override
     @ForceInline
     public Byte128Vector lanewise(Binary op, Vector<Byte> v) {
-        return (Byte128Vector) super.lanewiseTemplate(op, Byte128Mask.class, v);  // specialize
+        return (Byte128Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte256Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte256Vector.java
@@ -276,7 +276,7 @@ final class Byte256Vector extends ByteVector {
     @Override
     @ForceInline
     public Byte256Vector lanewise(Binary op, Vector<Byte> v) {
-        return (Byte256Vector) super.lanewiseTemplate(op, Byte256Mask.class, v);  // specialize
+        return (Byte256Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte512Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte512Vector.java
@@ -276,7 +276,7 @@ final class Byte512Vector extends ByteVector {
     @Override
     @ForceInline
     public Byte512Vector lanewise(Binary op, Vector<Byte> v) {
-        return (Byte512Vector) super.lanewiseTemplate(op, Byte512Mask.class, v);  // specialize
+        return (Byte512Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte64Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte64Vector.java
@@ -276,7 +276,7 @@ final class Byte64Vector extends ByteVector {
     @Override
     @ForceInline
     public Byte64Vector lanewise(Binary op, Vector<Byte> v) {
-        return (Byte64Vector) super.lanewiseTemplate(op, Byte64Mask.class, v);  // specialize
+        return (Byte64Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteMaxVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteMaxVector.java
@@ -276,7 +276,7 @@ final class ByteMaxVector extends ByteVector {
     @Override
     @ForceInline
     public ByteMaxVector lanewise(Binary op, Vector<Byte> v) {
-        return (ByteMaxVector) super.lanewiseTemplate(op, ByteMaxMask.class, v);  // specialize
+        return (ByteMaxVector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double128Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double128Vector.java
@@ -276,7 +276,7 @@ final class Double128Vector extends DoubleVector {
     @Override
     @ForceInline
     public Double128Vector lanewise(Binary op, Vector<Double> v) {
-        return (Double128Vector) super.lanewiseTemplate(op, Double128Mask.class, v);  // specialize
+        return (Double128Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double256Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double256Vector.java
@@ -276,7 +276,7 @@ final class Double256Vector extends DoubleVector {
     @Override
     @ForceInline
     public Double256Vector lanewise(Binary op, Vector<Double> v) {
-        return (Double256Vector) super.lanewiseTemplate(op, Double256Mask.class, v);  // specialize
+        return (Double256Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double512Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double512Vector.java
@@ -276,7 +276,7 @@ final class Double512Vector extends DoubleVector {
     @Override
     @ForceInline
     public Double512Vector lanewise(Binary op, Vector<Double> v) {
-        return (Double512Vector) super.lanewiseTemplate(op, Double512Mask.class, v);  // specialize
+        return (Double512Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double64Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Double64Vector.java
@@ -276,7 +276,7 @@ final class Double64Vector extends DoubleVector {
     @Override
     @ForceInline
     public Double64Vector lanewise(Binary op, Vector<Double> v) {
-        return (Double64Vector) super.lanewiseTemplate(op, Double64Mask.class, v);  // specialize
+        return (Double64Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/DoubleMaxVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/DoubleMaxVector.java
@@ -276,7 +276,7 @@ final class DoubleMaxVector extends DoubleVector {
     @Override
     @ForceInline
     public DoubleMaxVector lanewise(Binary op, Vector<Double> v) {
-        return (DoubleMaxVector) super.lanewiseTemplate(op, DoubleMaxMask.class, v);  // specialize
+        return (DoubleMaxVector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/DoubleVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/DoubleVector.java
@@ -625,7 +625,6 @@ public abstract class DoubleVector extends AbstractVector<Double> {
     @ForceInline
     final
     DoubleVector lanewiseTemplate(VectorOperators.Binary op,
-                                          Class<? extends VectorMask<Double>> maskClass,
                                           Vector<Double> v) {
         DoubleVector that = (DoubleVector) v;
         that.check(this);
@@ -646,7 +645,7 @@ public abstract class DoubleVector extends AbstractVector<Double> {
 
         int opc = opCode(op);
         return VectorSupport.binaryMaskedOp(
-            opc, getClass(), maskClass, double.class, length(),
+            opc, getClass(), null, double.class, length(),
             this, that, null,
             BIN_MASKED_IMPL.find(op, opc, DoubleVector::binaryOperations));
     }
@@ -786,8 +785,7 @@ public abstract class DoubleVector extends AbstractVector<Double> {
     DoubleVector lanewise(VectorOperators.Binary op,
                                   long e) {
         double e1 = (double) e;
-        if ((long)e1 != e
-            ) {
+        if ((long)e1 != e) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1);
@@ -808,8 +806,7 @@ public abstract class DoubleVector extends AbstractVector<Double> {
     DoubleVector lanewise(VectorOperators.Binary op,
                                   long e, VectorMask<Double> m) {
         double e1 = (double) e;
-        if ((long)e1 != e
-            ) {
+        if ((long)e1 != e) {
             vspecies().checkValue(e);  // for exception
         }
         return lanewise(op, e1, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Float128Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Float128Vector.java
@@ -276,7 +276,7 @@ final class Float128Vector extends FloatVector {
     @Override
     @ForceInline
     public Float128Vector lanewise(Binary op, Vector<Float> v) {
-        return (Float128Vector) super.lanewiseTemplate(op, Float128Mask.class, v);  // specialize
+        return (Float128Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Float256Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Float256Vector.java
@@ -276,7 +276,7 @@ final class Float256Vector extends FloatVector {
     @Override
     @ForceInline
     public Float256Vector lanewise(Binary op, Vector<Float> v) {
-        return (Float256Vector) super.lanewiseTemplate(op, Float256Mask.class, v);  // specialize
+        return (Float256Vector) super.lanewiseTemplate(op, v);  // specialize
     }
 
     @Override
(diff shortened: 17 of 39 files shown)


XiaohongGong:
> Hi Xiaohong,
> 
> Thanks a lot for this. We are almost there. I was hoping that this PR will only have the java and blend node creation in backend. No predicate register support, no new IR nodes. i.e. in this first patch Matcher::match_rule_supported_masked_vector() set as false for every architecture including SVE.
> 
> My thought for smooth progress steps is as follows:
> 1) Xiaohong commits *.java and blend node creation with Matcher::match_rule_supported_masked_vector() set as false for every architecture
> 2) Jatin commits mainline predicate register patch with x86 predicate registers
> 3) Xiaohong commits SVE predicate register support and basic new nodes creation
> 4) Both x86/sve work can then go on in parallel.
> 
> For steps 1) the code needs to be approved by Paul Sandoz and Vladimir Ivanov. This step is most straight forward, and we should be able to get it in quickly.
> For step 2) and 3) new node definition code needs to be approved by Vladimir Ivanov. Here step 2) has been already discussed extensively. Step 3) might need some discussion and back and forth.
> Step 4) we could request for guidance from Vladimir Ivanov as needed
> 
> Please let me know what you think of this or if you have alternate suggestion.
> 
> Best Regards,
> Sandhya

Hi @sviswa7 ,

Thanks for your detailed suggestion for the whole progress! The steps seem reasonable for me. I will remove the RA and new added mask IRs soon. BTW, please see my questions about step 1) and 3). 

> 1) Xiaohong commits *.java and blend node creation with Matcher::match_rule_supported_masked_vector() set as false for every architecture

I agree to add this part at the first patch. However, to easier review the masking support solution, I only added the support for masked binary and store APIs. Do you think it's needed to add all the same changes for other masked APIs (unary, ternary, reduction) together in the same PR? Or adding the missing parts after the solution is high-levelly reviewed?

> 3) Xiaohong commits SVE predicate register support and basic new nodes creation

I think @nsjian will create a separate patch for the SVE predicate register part that is based on Jatin's mainline predicate register patch first.  And I will commit the new added mask IR parts based on the that after then.  This part will not contain any backend implementations. So we could parallelly work on the AVX-512/SVE backend support finally as you suggested (Step 4).

Please let me know if any comments. Thanks!

Best Regards,
Xiaohong

XiaohongGong:
Hi, the scope of this PR is shortened to the Vector API java implementation and hotspot vector intrinsic changes. It adds two new hotspot vector intrinsic methods specially for the the masked binary/store, and the hotspot implementations for them.

The new added binary mask intrinsic is used both by masked and non-masked binary operations. And the mask argument is set to be `"null"` for non-masked version. The hotspot will generate different codes due to different mask values:

 1) If the mask is null, the normal no mask controlled codes are generated.
 2) If mask is non-null, the optimized predicated codes are generated for platforms that support mask feature (i.g. SVE/AVX-512).
    Otherwise, the vector blend codes are generated like before.

And the new added store masked intrinsic is used by masked vector store. The hotspot will generate the predicated vector store codes for platforms like SVE/AVX-512, or return to the default java implementation for other platforms.

Note that since no backend codes are added and the hotspot checking for the supported masked operations is set to false for all platforms currently, this patch will always generate the blend pattern as before.

All the backend implementations and other compiler changes will be in a separate patch.

Thanks,
Xiaohong Gong

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -674,91 +691,76 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
                 $abstractvectortype$ neg = that.lanewise(NEG);
                 $abstractvectortype$ hi = this.lanewise(LSHL, (op == ROR) ? neg : that);
                 $abstractvectortype$ lo = this.lanewise(LSHR, (op == ROR) ? that : neg);
-                return hi.lanewise(OR, lo);
+                return m != null ? blend(hi.lanewise(OR, lo), m) : hi.lanewise(OR, lo);
```
sviswa7, line 694: Checking for m!=null would cause boxing here.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -674,91 +691,76 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
...
             } else if (op == AND_NOT) {
                 // FIXME: Support this in the JIT.
                 that = that.lanewise(NOT);
                 op = AND;
             } else if (op == DIV) {
                 VectorMask<$Boxtype$> eqz = that.eq(($type$)0);
-                if (eqz.anyTrue()) {
-                    throw that.divZeroException();
+                if (m != null) {
```
sviswa7, line 701: Checking for m!=null would cause boxing here.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -823,7 +817,16 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
     $abstractvectortype$ lanewise(VectorOperators.Binary op,
                                   $type$ e,
                                   VectorMask<$Boxtype$> m) {
-        return blend(lanewise(op, e), m);
+#if[BITWISE]
+        if (opKind(op, VO_SHIFT) && ($type$)(int)e == e) {
+            $abstractvectortype$ shift = lanewiseShift(op, (int) e);
+            return m != null ? blend(shift, m) : shift;
```
sviswa7, line 823: Checking for m!=null would cause boxing here.

sviswa7:
Thanks a lot @XiaohongGong for this smaller patch to get us started.
@PaulSandoz Do the Java code changes here look good to you, please advice?
@iwanowww Does the C2 and VectorSupport.java related code changes look good to you, please advice?

src/hotspot/share/opto/vectornode.cpp
```diff
@@ -752,7 +752,16 @@ Node* StoreVectorMaskedNode::Ideal(PhaseGVN* phase, bool can_reshape) {
...
-  return NULL;
+  return StoreNode::Ideal(phase, can_reshape);
+}
+
+const TypeVect* StoreVectorMaskedNode::vect_type() const {
+  Node* value = in(MemNode::ValueIn);
+  if (value->bottom_type()->isa_vect() != NULL) {
+    return value->bottom_type()->is_vect();
+  }
+  // The value input is actually a BinaryNode, which has two inputs value and mask.
```
jatin-bhateja, line 763: BinaryNodes are created just before selection. In other cases a vector binary operation should be store's input value and its bottom_type() must be a valid vector type. Did you see any failure/issue for which this special handling for Binarynode is added?

jatin-bhateja (commented):
```
src/hotspot/share/opto/vectornode.cpp
  if (value->bottom_type()->isa_vect() != NULL) {
    return value->bottom_type()->is_vect();
  }
  // The value input is actually a BinaryNode, which has two inputs value and mask.
```

BinaryNodes are created just before selection. In other cases a vector binary operation should be store's input value and its bottom_type() must be a valid vector type. Did you see any failure/issue for which this special handling for Binarynode is added?

src/hotspot/share/opto/vectornode.cpp
XiaohongGong: Yes, this issue is caused due to the matcher puts the third and forth inputs into a BinaryNode before matching. See: https://github.com/openjdk/panama-vector/blob/vectorIntrinsics/src/hotspot/share/opto/matcher.cpp#L2348

The issue will happen when `vect_type()` is called after then (i.g. in the ad file).

XiaohongGong: I guess the latest code doesn't have this issue, due to https://github.com/openjdk/jdk/pull/2867. I will update the codes and have a test. Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
XiaohongGong: Yeah, that's true. How about handling the difference separately in the masked/non-masked `lanewise` method?

XiaohongGong: My initial thought is to add a new abstract method i.g. `lanewise0()`. It did the same transformations for special opcodes and then directly calls the intrinsic method `binaryMaskOp`. All the different handles for special ops are handled differently in the `lanewise` method. 
The codes might look like:
```
   /**
     * {@inheritDoc} <!--workaround-->
     * @see #lanewise(VectorOperators.Binary,byte)
     * @see #lanewise(VectorOperators.Binary,byte,VectorMask)
     */
    @ForceInline
    public final
    ByteVector lanewise(VectorOperators.Binary op,
                                  Vector<Byte> v) {
        ByteVector that = (ByteVector) v;
        that.check(this);
        if (op == ROR || op == ROL) { // FIXME: JIT should do this
            ByteVector neg = that.lanewise(NEG);
            ByteVector hi = this.lanewise(LSHL, (op == ROR) ? neg : that);
            ByteVector lo = this.lanewise(LSHR, (op == ROR) ? that : neg);
            return hi.lanewise(OR, lo);
        } else if (op == DIV) {
            VectorMask<Byte> eqz = that.eq((byte)0);
            if (eqz.anyTrue()) {
                throw that.divZeroException();
            }
        }
        return lanewise0(op, that, null);
    }

    /**
     * {@inheritDoc} <!--workaround-->
     * @see #lanewise(VectorOperators.Binary,byte,VectorMask)
     */
    @Override
    public final
    ByteVector lanewise(VectorOperators.Binary op,
                                  Vector<Byte> v,
                                  VectorMask<Byte> m) {
        ByteVector that = (ByteVector) v;
        that.check(this);
        if (op == ROR || op == ROL) {
            return blend(lanewise(op, v), m);
        } else if (op == DIV) {
            VectorMask<Byte> eqz = that.eq((byte)0);
            if (eqz.and(m).anyTrue()) {
                throw that.divZeroException();
            }
            // suppress div/0 exceptions in unset lanes
            that = that.lanewise(NOT, eqz);
        }
        return lanewise0(op, that, m);
    }

    protected abstract
    ByteVector lanewise0(VectorOperators.Binary op,
                         Vector<Byte> v,
                         VectorMask<Byte> m);

    @ForceInline
    final
    ByteVector lanewise0Template(VectorOperators.Binary op,
                                          Class<? extends VectorMask<Byte>> maskType,
                                          Vector<Byte> v, VectorMask<Byte> m) {
        ByteVector that = (ByteVector) v;
        if (opKind(op, VO_SPECIAL  | VO_SHIFT)) {
            if (op == FIRST_NONZERO) {
                // FIXME: Support this in the JIT.
                VectorMask<Byte> thisNZ
                        = this.viewAsIntegralLanes().compare(NE, (byte) 0);
                that = that.blend((byte) 0, thisNZ.cast(vspecies()));
                op = OR_UNCHECKED;
            }
            if (opKind(op, VO_SHIFT)) {
                // As per shift specification for Java, mask the shift count.
                // This allows the JIT to ignore some ISA details.
                that = that.lanewise(AND, SHIFT_MASK);
            }
            if (op == AND_NOT) {
                // FIXME: Support this in the JIT.
                that = that.lanewise(NOT);
                op = AND;
            }
        }

        int opc = opCode(op);
        return VectorSupport.binaryMaskOp(
            opc, getClass(), maskType, byte.class, length(),
            this, that, m,
            BIN_MASK_IMPL.find(op, opc, (opc_) -> {
              switch (opc_) {
                case VECTOR_OP_ADD: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a + b));
                case VECTOR_OP_SUB: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a - b));
                case VECTOR_OP_MUL: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a * b));
                case VECTOR_OP_DIV: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a / b));
                case VECTOR_OP_MAX: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)Math.max(a, b));
                case VECTOR_OP_MIN: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)Math.min(a, b));
                case VECTOR_OP_AND: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a & b));
                case VECTOR_OP_OR: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a | b));
                case VECTOR_OP_XOR: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, b) -> (byte)(a ^ b));
                case VECTOR_OP_LSHIFT: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, n) -> (byte)(a << n));
                case VECTOR_OP_RSHIFT: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, n) -> (byte)(a >> n));
                case VECTOR_OP_URSHIFT: return (v0, v1, vm) ->
                        v0.bOp(v1, vm, (i, a, n) -> (byte)((a & LSHR_SETUP_MASK) >>> n));
                default: return null;
                }}));
    }
```
The additional benefit for the masked DIV is that it only needs one "div zero checking" for the active elements with this change, while it checks twice with the original (`blend`) codes.  The original codes firstly check the active elements (calls by masked lanewise), and then check all the elements (calls by non-masked lanewise) which I think is duplicate.

Does it look ok for you? Thanks!

Best Regards,
Xiaohong

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
XiaohongGong: See the above solution. Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
XiaohongGong: Since I didn't add the masked intrinisic version for `broadcastInt` that is called by `lanewiseShift`. I think it can be fixed with the same solution for other binary `lanewise`. Once the masked `broadcastInt` is added, the code might look like:
```
   if (opKind(op, VO_SHIFT) && ($type$)(int)e == e) {
         return lanewiseShift(op, (int) e, m);
   }
```

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
sviswa7: This looks good to me. The ForceInline annotation needs to be added to the following method.
ByteVector lanewise(VectorOperators.Binary op,
                                  Vector<Byte> v,
                                  VectorMask<Byte> m)

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
sviswa7: ok.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
sviswa7: sounds good.

XiaohongGong:
Hi @sviswa7 , all your comments have been addressed! Would you mind having a look at it again? Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
```diff
@@ -591,16 +593,66 @@ IntVector lanewise(VectorOperators.Unary op,
...
         IntVector that = (IntVector) v;
         that.check(this);
-        if (opKind(op, VO_SPECIAL  | VO_SHIFT)) {
+        if (op == DIV) {
+            VectorMask<Integer> eqz = that.eq((int)0);
+            if (eqz.and(m).anyTrue()) {
+                throw that.divZeroException();
+            }
+            // suppress div/0 exceptions in unset lanes
+            that = that.lanewise(NOT, eqz);
```
PaulSandoz, line 640: I guess this is required because we don't know how the intrinsic will support masking? How can a  div/0 exception can occur?

PaulSandoz (commented):
The changes to Java code generally look ok. @iwanowww will know more than I with regards to how this all behaves with C2 e.g. using `null` or some other sentinel, such as a constant mask with all bits set?

There is an awkward set of differences between non-mask and mask for ROR etc. Although most of that would go away if support is in C2. Possibly leaving subtle differences for DIV. I imagine adjustment of templates will be similar.

I am tempted to have a template method for both non-mask and mask. Thus specialized code (that before calling the intrinsic) may be somewhat duplicated before calling `binaryMaskOp`, using a constant for the op  -> lambda function, and passing in null or the mask value.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
XiaohongGong: Thanks for looking at this PR @PaulSandoz ! Yes, this is actually required for masked division operation. I think it has to make sure no div/0 exception happens during hotspot compiling the intrinsic? Am I missing something? Thanks!

XiaohongGong:
> The changes to Java code generally look ok. @iwanowww will know more than I with regards to how this all behaves with C2 e.g. using `null` or some other sentinel, such as a constant mask with all bits set?

Hi @PaulSandoz , thanks for the comments! Yeah, we can also use a `allTrue` constant mask for the non-mask APIs. This seems reasonable, with just a shortage that the compiler needs to generate more codes to check whether the mask is all true or not. Considering the performance, I didn't use this way. Hope @iwanowww can give more advice for this. Thanks!

> There is an awkward set of differences between non-mask and mask for ROR etc. Although most of that would go away if support is in C2. Possibly leaving subtle differences for DIV. I imagine adjustment of other templates will be similar.

Yes, the adjustment of other templates will be similar.

> I am tempted to have a template method for both non-mask and mask. Thus specialized code (that before calling the intrinsic) may be somewhat duplicated before calling `binaryMaskOp`, using a constant for the op -> lambda function, and passing in null or the mask value.

Yeah, that's why I added the `lanewise0` and `lanewise0Template` methods.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
PaulSandoz: Did you observe any issues without this? I understand how such exceptions can occur when compiling the instrinsic, but could be missing something.

PaulSandoz:
> > I am tempted to have a template method for both non-mask and mask. Thus specialized code (that before calling the intrinsic) may be somewhat duplicated before calling `binaryMaskOp`, using a constant for the op -> lambda function, and passing in null or the mask value.
> 
> Yeah, that's why I added the `lanewise0` and `lanewise0Template` methods.

I am suggesting to separate out templates for non-masked and masked, similar to the current pattern. That would result in some duplication for the special cases, but i think makes it easier to reason about in one place, rather than being somewhat spread out. Perhaps easier to explain as code. We should just iterate on the branch.

PaulSandoz (approved):

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/IntVector.java
XiaohongGong: I tried to remove `that = that.lanewise(NOT, eqz)`, but not issues are found. I think this is because currently the hotspot doesn't implement the intrinsic for integer division. And it seems this is only added for the division with integer types which might not be useful. And there are no need with any zero checks for `float/double` types in JAVA.

XiaohongGong:
> > > I am tempted to have a template method for both non-mask and mask. Thus specialized code (that before calling the intrinsic) may be somewhat duplicated before calling `binaryMaskOp`, using a constant for the op -> lambda function, and passing in null or the mask value.
> > 
> > 
> > Yeah, that's why I added the `lanewise0` and `lanewise0Template` methods.
> 
> I am suggesting to separate out templates for non-masked and masked, similar to the current pattern. That would result in some duplication for the special cases, but i think makes it easier to reason about in one place, rather than being somewhat spread out. Perhaps easier to explain as code. We should just iterate on the branch.

Thanks for your more details! I think I misunderstood your point before. I will think about your suggestion. Thanks so much!

src/hotspot/share/opto/library_call.cpp
```diff
@@ -633,6 +633,8 @@ bool LibraryCallKit::try_to_inline(int predicate) {
     return inline_vector_nary_operation(2);
```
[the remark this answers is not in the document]
XiaohongGong: Yes, this will be merge with the masked version in future. Since I didn't add the codes for unary/ternary operations, I choose this to be leaved as it is. We can remove it after all opcodes for unary/ternary are supported.

src/hotspot/share/classfile/vmIntrinsics.hpp
```diff
@@ -818,6 +818,11 @@ class methodHandle;
                                        "Ljava/util/function/BiFunction;)Ljava/lang/Object;")                                                   \
    do_name(vector_binary_op_name,     "binaryOp")                                                                                              \
```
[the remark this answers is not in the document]
XiaohongGong: I think this is not used anymore. I didn't do any cleanup, since currently unary/ternary still use the old version. We can do the cleanup together with unary/ternary in future.

sviswa7 (approved):

XiaohongGong:
Thanks for your review @sviswa7 @PaulSandoz @jatin-bhateja! Any comments from this PR @iwanowww ?

jatin-bhateja (approved):
Would prefer suggested  cleanup to be done along with this patch instead of pending it to next one.
Rest of the changes looks ok.

XiaohongGong:
> Would prefer suggested cleanup to be done along with this patch instead of pending it to next one.
> Rest of the changes looks ok.

I‘m sorry that I mislead you before. Currently the `binaryOp` cannot be cleanup since it is used by the vector mask `and/or/xor`. I will try to look at whether it can be replaced with `binaryMaskOp` as well. If so, we can have a separate patch for this. WDYT? Thanks!

jatin-bhateja:
> > Would prefer suggested cleanup to be done along with this patch instead of pending it to next one.
> > Rest of the changes looks ok.
> 
> I‘m sorry that I mislead you before. Currently the `binaryOp` cannot be cleanup since it is used by the vector mask `and/or/xor`. I will try to look at whether it can be replaced with `binaryMaskOp` as well. If so, we can have a separate patch for this. WDYT? Thanks!

Since for a non-masked operation null is being passed to binaryMaskOp so VectorMask.and/or/xor/not APIs seems no different.

XiaohongGong:
> > > Would prefer suggested cleanup to be done along with this patch instead of pending it to next one.
> > > Rest of the changes looks ok.
> > 
> > 
> > I‘m sorry that I mislead you before. Currently the `binaryOp` cannot be cleanup since it is used by the vector mask `and/or/xor`. I will try to look at whether it can be replaced with `binaryMaskOp` as well. If so, we can have a separate patch for this. WDYT? Thanks!
> 
> Since for a non-masked operation null is being passed to binaryMaskOp so VectorMask.and/or/xor/not APIs seems no different.

Yes, I think so. I will prepare a separate patch to update all these codes. Thanks!

src/hotspot/share/opto/vectornode.cpp
```diff
@@ -752,7 +752,7 @@ Node* StoreVectorMaskedNode::Ideal(PhaseGVN* phase, bool can_reshape) {
       return phase->transform(new StoreVectorNode(ctr, mem, adr, adr_type(), val));
     }
   }
-  return NULL;
+  return StoreNode::Ideal(phase, can_reshape);
```
iwanowww, line 755: Can you elaborate, please, why this change is needed?

src/hotspot/share/opto/library_call.hpp
```diff
@@ -310,10 +310,12 @@ class LibraryCallKit : public GraphKit {
 
   // Vector API support
   bool inline_vector_nary_operation(int n);
+  bool inline_vector_nary_mask_operation(int n);
```
iwanowww, line 313: Should it be named `inline_vector_nary_masked_operation()`?

src/hotspot/share/opto/matcher.hpp
```diff
@@ -317,6 +317,8 @@ class Matcher : public PhaseTransform {
   // e.g. Op_ vector nodes and other intrinsics while guarding with vlen
   static const bool match_rule_supported_vector(int opcode, int vlen, BasicType bt);
 
+  static const bool match_rule_supported_masked_vector(int opcode, int vlen, BasicType bt);
```
iwanowww, line 320: IMO `Matcher::match_rule_supported_vector_masked()` is a better option.

src/hotspot/share/classfile/vmIntrinsics.hpp
```diff
@@ -818,6 +818,11 @@ class methodHandle;
...
                                                                                                                                                \
+  do_intrinsic(_VectorBinaryMaskOp, jdk_internal_vm_vector_VectorSupport, vector_binary_mask_op_name, vector_binary_mask_op_sig, F_S)          \
```
iwanowww, line 821: Should it be named `VectorBinaryMaskedOp`?

iwanowww (commented):
I have one request: `m == null` should be part of agreement between JVM and JDK and hidden from users. It means that all non-masked operations (when/if they are migrated to masked intrinsics) should pass `null` constant as a mask and all masked operations should perform the null check to ensure mask value is always non-null and clearly communicate that to JITs.

I think it is important because it is a prerequisite for successful intrinsification: unless C2 can't definitely say whether the mask is `null` or non-`null`, it can't make the decision how to implement the operation. So, we should be cautious about possible inlining and profile pollution effects and ensure that all usages of `VectorSupport::binaryMaskedOp()/storeMasked()` provide enough information for C2 to detect proper operation mode.

XiaohongGong:
> I have one request: `m == null` should be part of internal agreement between JVM and JDK and completely hidden from users. It means that all non-masked operations (when/if they are migrated to masked intrinsics) should pass `null` constant as a mask and all masked operations should perform the null check to ensure mask value is always non-null and clearly communicate that to JITs.

Thanks for your comments @iwanowww ! This is a really great question that I also considered it before. Yes, I think it has to guarantee the mask value is non-null for masked operations, so that the JIT can clearly know how to generate right codes. Since the `m == null` check in API implementation could block the vector box elimination, I think we could add a check (e.g. `this.check(m);`) like what have done for two normal vectors for the masked `lanewise` APIs. Although this is mainly used to check the species of two vectors, it can expose the `null` value of mask at the beginning of API. Sounds ok? Thanks!

src/hotspot/share/opto/vectornode.cpp
XiaohongGong: Oh, I just noticed that the `StoreVectorMaskedNode` is extended from `StoreNode`, which has some optimizations in `StoreNode::Ideal`. I'm worried that overriding the `StoreNode::Ideal` for `StoreVectorMaskedNode` might miss the optimizations in `StoreNode::Ideal`. I will reconsider this in more details. Thanks!

iwanowww:
It's better to ask @PaulSandoz about how to better shape implementation code on JDK side.

IMO having an explicit null check (e.g., `Objects.requireNonNull()`) as part of argument validation in all public methods for masked operations should be a bare minimum. We already rely on the ability to propagate all the important information from the call site down to the intrinsic call.

src/hotspot/share/opto/vectornode.cpp
iwanowww: It makes sense to handle it separately and fix it in the mainline.

jatin-bhateja:
> > I have one request: `m == null` should be part of internal agreement between JVM and JDK and completely hidden from users. It means that all non-masked operations (when/if they are migrated to masked intrinsics) should pass `null` constant as a mask and all masked operations should perform the null check to ensure mask value is always non-null and clearly communicate that to JITs.
> 
> Thanks for your comments @iwanowww ! This is a really great question that I also considered it before. Yes, I think it has to guarantee the mask value is non-null for masked operations, so that the JIT can clearly know how to generate right codes. Since the `m == null` check in API implementation could block the vector box elimination, I think we could add a check (e.g. `this.check(m);`) like what have done for two normal vectors for the masked `lanewise` APIs. Although this is mainly used to check the species of two vectors, it can expose the `null` value of mask at the beginning of API. Sounds ok? Thanks!

Passing an extra primitive flag whose value is true for masked and false otherwise could be a safe check to skip intrinsification if mask argument is null for a masked vector operation. As @iwanowww suggested @PaulSandoz should be able to guide us here.

PaulSandoz:
The simplest approach for mask accepting operations may be just to call `m.getClass()`, to induce an implicit `null` check?

Separating out the templates for mask/non-mask (as suggested previously) where the template passes a non-`null` (and implicitly `null`-checked) mask value, or explicitly `null` for no mask, to the intrinsic may be easier to manage, and the code is clear in the intent (the explicit `null` argument is declared at the intrinsic call site).

Otherwise, if the mask/non-mask template is unified it may be simpler for Java code to leverage an `ALL_TRUE` constant. However, that might be harder for HS to deal with to detect the mask value is of the required constant.

Either way i think we can avoid an extra boolean flag and drive the signal from the mask value itself.

src/hotspot/share/opto/vectornode.cpp
XiaohongGong: It's ok for me. I will revert this change. Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
```diff
@@ -625,34 +641,15 @@ ByteVector lanewise(VectorOperators.Binary op,
...
-            VectorMask<Byte> eqz = that.eq((byte)0);
-            if (eqz.and(m).anyTrue()) {
-                throw that.divZeroException();
-            }
-            // suppress div/0 exceptions in unset lanes
-            that = that.lanewise(NOT, eqz);
+        VectorSpecies<Byte> maskSpecies =
+            ((jdk.incubator.vector.VectorMask<Byte>) m).vectorSpecies();
+        if (maskSpecies != vspecies()) {
+            throw AbstractSpecies.checkFailed(maskSpecies, vspecies());
```
XiaohongGong, line 649: @iwanowww @PaulSandoz , the new commit add a type check to the mask value in the masked `lanewise` API. This could raise an implicit null check for the mask value. I tried to use `m.getClass()` at first, but I didn't find a reasonable way to use it in this API. Does this change make sense for you? Thanks!

XiaohongGong:
> > > > I am tempted to have a template method for both non-mask and mask. Thus specialized code (that before calling the intrinsic) may be somewhat duplicated before calling `binaryMaskOp`, using a constant for the op -> lambda function, and passing in null or the mask value.
> > > 
> > > 
> > > Yeah, that's why I added the `lanewise0` and `lanewise0Template` methods.
> > 
> > 
> > I am suggesting to separate out templates for non-masked and masked, similar to the current pattern. That would result in some duplication for the special cases, but i think makes it easier to reason about in one place, rather than being somewhat spread out. Perhaps easier to explain as code. We should just iterate on the branch.
> 
> Thanks for your more details! I think I misunderstood your point before. I will think about your suggestion. Thanks so much!

Hi @PaulSandoz , the latest commit changes something according to your suggestion. It totally separateS the special handles for the special cases for non-masked and masked version, although finally they still use one template `lanewise0Template`. Could you please take a look again? If it matches what you suggested above, do you think it need to separate out the template methods further? Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
iwanowww: Yes, it looks fine to me. 

I suggest to factor the check into a helper method (`AbstractVector::check(VectorMask m)`) and call it next to the check on `Vector`.

PaulSandoz:
@XiaohongGong See the following diff which more accurately depicts what i was suggesting:
https://gist.github.com/PaulSandoz/d93fa2a78c3c7be5fdb2d714371639ce
i.e. we follow the exiting template pattern more directly in both the non-mask and mask.
Perhaps we find later on we can consolidate, but right now i find that pattern easier to follow and it is very clear on the null contract for no mask.

XiaohongGong:
> @XiaohongGong See the following diff which more accurately depicts what i was suggesting:
> https://gist.github.com/PaulSandoz/d93fa2a78c3c7be5fdb2d714371639ce
> i.e. we follow the exiting template pattern more directly in both the non-mask and mask.
> Perhaps we find later on we can consolidate, but right now i find that pattern easier to follow and it is very clear on the null contract for no mask.

Thanks for your suggestion! I also tried this way before the current version. However, the main concern for me is the species checking for vector mask.  See the codes:
```
    @Override
    @ForceInline
    public final Int128Vector lanewise(Binary op, Vector<Integer> v, VectorMask<Integer> m) {
        return (Int128Vector) super.lanewiseTemplate(op, Int128Mask.class, v, (Int128Mask) m);  // specialize
    }
```
The override `lanewise` method must cast the mask value to `Int128Mask` before it calls the super class's `lanewiseTemplate` which does the species checking later. It seems more reasonable for me to do the species checking before doing the type casting. That means the species checking must be done in the `lanewise` method while not `lanewiseTemplate`. Add considering adding the species checking for each subclass's `lanewise` seems duplicate, finally I change the codes to the current style.

Wanghuang-Huawei:
Look good to me. Thank you for your work.

Wanghuang-Huawei (approved):

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
XiaohongGong: Thanks! I will update the codes according to your suggestion!

XiaohongGong:
Hi @iwanowww , I just updated the codes according to your comment. Would you mind having a look again? Thanks so much!

PaulSandoz:
> > @XiaohongGong See the following diff which more accurately depicts what i was suggesting:
> > https://gist.github.com/PaulSandoz/d93fa2a78c3c7be5fdb2d714371639ce
> > i.e. we follow the exiting template pattern more directly in both the non-mask and mask.
> > Perhaps we find later on we can consolidate, but right now i find that pattern easier to follow and it is very clear on the null contract for no mask.
> 
> Thanks for your suggestion! I also tried this way before the current version. However, the main concern for me is the species checking for vector mask. See the codes:
> 
> ```
>     @Override
>     @ForceInline
>     public final Int128Vector lanewise(Binary op, Vector<Integer> v, VectorMask<Integer> m) {
>         return (Int128Vector) super.lanewiseTemplate(op, Int128Mask.class, v, (Int128Mask) m);  // specialize
>     }
> ```
> 
> The override `lanewise` method must cast the mask value to `Int128Mask` before it calls the super class's `lanewiseTemplate` which does the species checking later. It seems more reasonable for me to do the species checking before doing the type casting. That means the species checking must be done in the `lanewise` method while not `lanewiseTemplate`. Add considering adding the species checking for each subclass's `lanewise` seems duplicate, finally I change the codes to the current style.

Can you explain more about why it is a concern? Is there some limitation you are observing with HotSpot? (I could be forgetting something, this is a long thread!)

I wonder if the cast is now required given a form of species check. Note that the argument vector (`v` or `that`) is not cast, although there is an equality check on classes in `sameSpecies`:

```java
    @ForceInline
    @SuppressWarnings("unchecked")
    /*package-private*/ final
    <F> AbstractVector<F> check(Vector<F> other) {
        if (!sameSpecies(other)) {
            throw AbstractSpecies.checkFailed(this, other);
        }
        return (AbstractVector<F>) this;
    }

    @ForceInline
    private boolean sameSpecies(Vector<?> other) {
        // It's simpler and faster to do a class check.
        boolean same = (this.getClass() == other.getClass());
        // Make sure it works, too!
        assert(same == (this.species() == other.species())) : same;
        return same;
    }
```

Can we follow a similar pattern e.g. `maskClass == m.getClass()`? Perhaps with check on `AbstractMask` and do `m.check(maskClass, v);` to induce an NPE before checking?:
```java
@ForceInline
@SuppressWarnings("unchecked")
/*package-private*/ final
<F> AbstractMask<F> check(Class<? extends VectorMask<F>> maskClass, Vector<F> other) {
    if (!sameSpecies(maskClass, other)) {
        throw AbstractSpecies.checkFailed(this, other);
    }
    return (AbstractMask<F>) this;
}

@ForceInline
<F> boolean sameSpecies(Class<? extends VectorMask<F>> maskClass, Vector<F> other) {
    same = this.getClass() == maskClass;
    assert (same == (vectorSpecies() == v.species())) : same;
    return same;
}
```

XiaohongGong:
@PaulSandoz , thanks for your comments! It can work well without the casting for mask. I'v updated the codes and separate the `lanewiseTemplate ` for both masked and non-masked operations. Thanks!

@iwanowww , any other comments about this PR? Is it possible to make this PR go into this branch so that we can start the next step for the masking support?  Thanks so much!

PaulSandoz:
> @PaulSandoz , thanks for your comments! It can work well without the casting for mask. I'v updated the codes and separate the `lanewiseTemplate ` for both masked and non-masked operations. Thanks!
> 

Very good!

PaulSandoz (approved):

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -376,6 +376,226 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
...
+        operation = VectorNode::make(sopc, opd1, opd2, opd3, vt);
+        break;
+      }
+      default: fatal("unsupported arity: %d", n);
+    }
+  }
+
+  if (is_masked_op && mask != NULL) {
+    if (use_predicate) {
+      // TODO: add predicate implementation for masked operation.
```
iwanowww, line 585: Does it make sense to fail the intrinsification or fail-fast with a crash here until the support is there?

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -376,6 +376,226 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
...
+    return false; // operation not supported
+  }
+  ciKlass* vbox_klass = vector_klass->const_oop()->as_instance()->java_lang_Class_klass();
+  const TypeInstPtr* vbox_type = TypeInstPtr::make_exact(TypePtr::NotNull, vbox_klass);
+
+  // "argument(n + 5)" should be the mask object. We assume it is "null" when no mask
+  // is used to control this operation.
+  bool is_masked_op = gvn().type(argument(n + 5)) != TypePtr::NULL_PTR;
+  if (is_vector_mask(vbox_klass)) {
+    assert(!is_masked_op, "mask operations do not need mask to control");
```
iwanowww, line 453: Another invariant to check here is mask input has to be either "always null" or "never null".

Currently, it will be checked when mask is unboxed, but it makes sense to make the check evident.

iwanowww (approved):
Overall, looks good. 

Just a couple of minor comments.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -4173,6 +4238,22 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
                       (arr_, off_, i, e) -> arr_[off_+i] = e));
     }
 
+    abstract
+    void intoArray0($type$[] a, int offset, VectorMask<$Boxtype$> m);
+    @ForceInline
+    final
+    <M extends VectorMask<$Boxtype$>>
+    void intoArray0Template(Class<M> maskClass, $type$[] a, int offset, M m) {
+        $Type$Species vsp = vspecies();
```
PaulSandoz, line 4247: Need `m.check(maskClass, this)`?

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -645,9 +647,11 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
     @ForceInline
     final
     $abstractvectortype$ lanewiseTemplate(VectorOperators.Binary op,
+                                          Class<? extends VectorMask<$Boxtype$>> maskClass,
                                           Vector<$Boxtype$> v) {
```
PaulSandoz, line 651:
```suggestion
    $abstractvectortype$ lanewiseTemplate(VectorOperators.Binary op,
                                          Vector<$Boxtype$> v) {
```
Remove unused argument by passing `null` to the intrinsic for the `maskClass` argument?

PaulSandoz (commented):
I left a few minor comments that could be addressed after integration of this PR.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
XiaohongGong: I think there is no similar issue with masked binary operations if missing this check here. There is implicit null check and type cast check for `m` before it calls `intoArray0`:
```
void intoArray(byte[] a, int offset,
                   VectorMask<Byte> m) {
        if (m.allTrue()) {
            intoArray(a, offset);
        } else {
            ByteSpecies vsp = vspecies();
            checkMaskFromIndexSize(offset, vsp, m, 1, a.length);
            intoArray0(a, offset, m);
        }
    }
```
The type cast exception will be thrown out inside `checkMaskFromIndexSize(offset, vsp, m, 1, a.length);` if the species are not matched. And we can see there is `m.allTrue()` at the begining of `intoArray`.

Anyway, adding this check is also ok for me. I can add it if adding the check looks better for you.

XiaohongGong:
@iwanowww @PaulSandoz all your significant comments have been addressed! Could you please have a look again? Thanks so much!

iwanowww (approved):
Looks good.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
PaulSandoz: > The type cast exception will be thrown out inside `checkMaskFromIndexSize(offset, vsp, m, 1, a.length);` if the species are not matched. And we can see there is `m.allTrue()` at the begining of `intoArray`.
> 
> Anyway, adding this check is also ok for me. I can add it if adding the check looks better for you.

Let's revisit after you integrate. We can iterate on this with additional PRs.

XiaohongGong: Thanks!

XiaohongGong:
Thanks for your review @sviswa7 @PaulSandoz @jatin-bhateja @iwanowww @Wanghuang-Huawei !

PR #65, opened 2021-04-13
8265109: SVE predicate register allocation support for vectorIntrinsics

nsjian:
This is the SVE predicate register allocation part for Vector API masking support. Since we don't have sufficent usages to test it in
jdk/jdk, and based on @sviswa7 's proposal [1], we would like to put it to vectorIntrinsics+mask first and get tested here with further predicate-based vector masking support, though so far no issue found with Xiaohong's initial Vector API masking support.

[1] https://mail.openjdk.java.net/pipermail/panama-dev/2021-April/012920.html

---------
### Progress

### Issue
 * [JDK-8265109](https://bugs.openjdk.java.net/browse/JDK-8265109): SVE predicate register allocation support for vectorIntrinsics

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/65/head:pull/65` \
`$ git checkout pull/65`

`$ git checkout pull/65` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/65/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 65`

`$ git pr show -t 65`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/65.diff">https://git.openjdk.java.net/panama-vector/pull/65.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/adlc/formssel.cpp
+++ b/src/hotspot/share/adlc/formssel.cpp
@@ -2275,7 +2275,7 @@ bool OperandForm::is_bound_register() const {
   if (strcmp(name, "RegD") == 0) size = 2;
   if (strcmp(name, "RegL") == 0) size = 2;
   if (strcmp(name, "RegN") == 0) size = 1;
-  if (strcmp(name, "RegVectMask") == 0) size = 1;
+  if (strcmp(name, "RegVectMask") == 0) size = globalAD->get_preproc_def("AARCH64") ? 1 : 2;
   if (strcmp(name, "VecX") == 0) size = 4;
   if (strcmp(name, "VecY") == 0) size = 8;
   if (strcmp(name, "VecZ") == 0) size = 16;


src/hotspot/share/adlc/formssel.cpp
```diff
@@ -2275,6 +2275,7 @@ bool OperandForm::is_bound_register() const {
   if (strcmp(name, "RegD") == 0) size = 2;
   if (strcmp(name, "RegL") == 0) size = 2;
   if (strcmp(name, "RegN") == 0) size = 1;
+  if (strcmp(name, "RegVectMask") == 0) size = 1;
```
sviswa7, line 2278: This might not work across architectures. @jatin-bhateja what are your thoughts on this?

src/hotspot/share/opto/chaitin.cpp
```diff
@@ -1371,6 +1389,11 @@ static OptoReg::Name find_first_set(LRG &lrg, RegMask mask, int chunk) {
         }
       }
       return OptoReg::Bad; // will cause chunk change, and retry next chunk
+    } else if (lrg._is_predicate) {
```
sviswa7, line 1392: lrg._is_predicate is set to 1 for all architectures, so using scalable_reg_slots on this path doesn't look correct.

nsjian: This is inside lrg.is_scalable() condition of L1362, which handles scalable predicate register specifically. I think for non-scalable arch, the normal path of regmask handling should be fine.

sviswa7: I missed that. Thanks, this looks good.

src/hotspot/share/adlc/formssel.cpp
jatin-bhateja: It doesn't have much usage apart from semantic checking for bound registers. 
But for correct ness a bounded opmask reg class for x86 has two registers.
Could not locate bounded reg class definitions for SVE predicate registers in aarch64.ad

nsjian: Yes, aarch64 does not need this for now. I can either revert this change (leaving it to you for future update if required) or update to size 2 for x86. Which do you prefer?

jatin-bhateja: I can see ADLC does parse the preprocessors in source sections, by adding a new #define SVE_ENABLE  1 in aarch64_sve.ad  and checking for the same similar to following handling may allow us to set size = 1 for SVE else 2.
https://github.com/openjdk/jdk/blob/master/src/hotspot/share/adlc/formssel.cpp#L2279

nsjian: That sounds good to me. Thanks @jatin-bhateja! Addressed in the new commit. Please let me know whether there are any other concerns.

jatin-bhateja: Thanks @nsjian , since adlc already consumes macros passed though command line your change looks good.

sviswa7 (approved):

nsjian:
Thank you @jatin-bhateja @sviswa7 for the review! So, integrate to the branch to unblock future work.

/integrate

PR #70, opened 2021-04-17
8265321: Add Rearrange nodes implementation for Arm SVE

Wanghuang-Huawei:
* Add Rearrange nodes implementation for Arm SVE, like rearrangeB/I/S/L
* add `sve_tbl` , which reads each element of the second source (index) vector and uses its value to select an indexed element from the first source (table) vector, and places the indexed table element in the destination vector element corresponding to the index vector element. If an index value is greater than or equal to the number of vector elements then it places zero in the corresponding destination vector element.[1]

[1]  https://developer.arm.com/documentation/ddi0596/2020-12/SVE-Instructions/TBL--Programmable-table-lookup-in-single-vector-table-?lang=en

---------
### Progress

### Issue
 * [JDK-8265321](https://bugs.openjdk.java.net/browse/JDK-8265321): Add Rearrange nodes implementation for Arm SVE

### Reviewers

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/70.diff">https://git.openjdk.java.net/panama-vector/pull/70.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -2411,6 +2411,16 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType 
     return false;
   }
   if (UseSVE > 0) {
+    switch (opcode) {
+      case Op_VectorLoadShuffle:
+      case Op_VectorRearrange:
+        if (vlen < 4) {
+          return false;
+        }
+        break;
+      default:
+        break;
+    }
     return op_sve_supported(opcode, vlen, bt);
   } else { // NEON
     // Special cases
--- a/src/hotspot/cpu/aarch64/aarch64_sve.ad
+++ b/src/hotspot/cpu/aarch64/aarch64_sve.ad
@@ -4015,7 +4015,7 @@ instruct rearrangeS(vReg dst, vReg src, vReg shuffle)
             n->bottom_type()->is_vect()->element_basic_type() == T_SHORT);
   match(Set dst (VectorRearrange src shuffle));
   ins_cost(SVE_COST);
-  format %{ "sve_tbl $dst, H, $src, $shuffle\t# vector rearrange (H)" %}
+  format %{ "sve_tbl $dst, H, $src, $shuffle\t# vector rearrange (S)" %}
   ins_encode %{
     __ sve_tbl(as_FloatRegister($dst$$reg), __ H,
                as_FloatRegister($src$$reg), as_FloatRegister($shuffle$$reg));
@@ -4029,7 +4029,7 @@ instruct rearrangeI(vReg dst, vReg src, vReg shuffle)
             n->bottom_type()->is_vect()->element_basic_type() == T_INT);
   match(Set dst (VectorRearrange src shuffle));
   ins_cost(SVE_COST);
-  format %{ "sve_tbl $dst, S, $src, $shuffle\t# vector rearrange (S)" %}
+  format %{ "sve_tbl $dst, S, $src, $shuffle\t# vector rearrange (I)" %}
   ins_encode %{
     __ sve_tbl(as_FloatRegister($dst$$reg), __ S,
                as_FloatRegister($src$$reg), as_FloatRegister($shuffle$$reg));
@@ -4043,7 +4043,7 @@ instruct rearrangeF(vReg dst, vReg src, vReg shuffle)
             n->bottom_type()->is_vect()->element_basic_type() == T_FLOAT);
   match(Set dst (VectorRearrange src shuffle));
   ins_cost(SVE_COST);
-  format %{ "sve_tbl $dst, S, $src, $shuffle\t# vector rearrange (S)" %}
+  format %{ "sve_tbl $dst, S, $src, $shuffle\t# vector rearrange (F)" %}
   ins_encode %{
     __ sve_tbl(as_FloatRegister($dst$$reg), __ S,
                as_FloatRegister($src$$reg), as_FloatRegister($shuffle$$reg));
@@ -4057,7 +4057,7 @@ instruct rearrangeL(vReg dst, vReg src, vReg shuffle)
             n->bottom_type()->is_vect()->element_basic_type() == T_LONG);
   match(Set dst (VectorRearrange src shuffle));
   ins_cost(SVE_COST);
-  format %{ "sve_tbl $dst, D, $src, $shuffle\t# vector rearrange (D)" %}
+  format %{ "sve_tbl $dst, D, $src, $shuffle\t# vector rearrange (L)" %}
   ins_encode %{
     __ sve_tbl(as_FloatRegister($dst$$reg), __ D,
                as_FloatRegister($src$$reg), as_FloatRegister($shuffle$$reg));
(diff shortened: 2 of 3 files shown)


src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -3993,3 +3992,89 @@ instruct loadshuffleL(vReg dst, vReg src)
...
+            n->bottom_type()->is_vect()->element_basic_type() == T_DOUBLE);
+  match(Set dst (VectorRearrange src shuffle));
+  ins_cost(SVE_COST);
+  format %{ "sve_tbl $dst, D, $src, $shuffle\t# vector rearrange (D)" %}
+  ins_encode %{
+    __ sve_tbl(as_FloatRegister($dst$$reg), __ D,
+               as_FloatRegister($src$$reg), as_FloatRegister($shuffle$$reg));
+  %}
+  ins_pipe(pipe_slow);
+%}
```
XiaohongGong, line 4080: At least merge `long/double` and `float/int` ? The format looks the same.

Wanghuang-Huawei: I think it's my fault. The comment is `L` for `rearrangeL` instead of `D`. I will fix it in my next patch.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -229,7 +229,6 @@ source %{
       case Op_LoadVectorGather:
       case Op_StoreVectorScatter:
       case Op_VectorLoadConst:
-      case Op_VectorRearrange:
```
XiaohongGong, line 232: The min vector length for `VectorRearrange/VecorLoadShuffle` is 4, which decides the min vector length for byte is 4. However, currently it doesn't have such size limitation for SVE. This makes the` rearrange` tests of `Double128VectorTests/Long128VectorTests ` crash when the compiler tries to make the byte vector type with length 2. Could you please fix this issue together with this patch? Thanks so much!

src/hotspot/cpu/aarch64/aarch64_sve.ad
nsjian: I also prefer merging them instead of having two separate rules. From the instruction point of view, there's only D, not L.

Wanghuang-Huawei: > I also prefer merging them instead of having two separate rules. From the instruction point of view, there's only D, not L.

Thank you for your review.

* I think that the comment is the node's comment (or rule's comment)  instead of single instruction's comment. This rule is `rearrangeL` which means rearranging long type so I think that the comment should be `L` instead of `D`.
* In `aarch64_neon.ad`, the comments are `B``I``F``D` and so on.

Wanghuang-Huawei: > > * I think that the comment is the node's comment (or rule's comment)  instead of single instruction's comment. This rule is `rearrangeL` which means rearranging long type so I think that the comment should be `L` instead of `D`.
> > * In `aarch64_neon.ad`, the comments are ` B`I`F`D `` and so on.
> 
> I assume you mean comment in format section. In aarch64_neon, some are S while some are F, but I don't think there's a need to generate two rules just for different format comments. If there's any, I would suggest to fix them as well.

Of course, in some case `L` and `D` can be zipped in one rule. However, in some case, `L` and `D` is difference. `D` is float type.  That is what I mentioned before `the comment is for the whole rule`.  In the comment, we should show the data type. ;-)

nsjian: Some cases are different, but I don't think we need to distinguish floating point data types for this operation as they are actually the same thing and generate the same code.

XiaohongGong: Agree with @nsjian . I prefer to merge all these rules into one as we have discussed too much times before, as there is no instruction difference for different types.  Anyway, we can have a separate patch to unify the styles in ad file.

Wanghuang-Huawei: > Some cases are different, but I don't think we need to distinguish floating point data types for this operation as they are actually the same thing and generate the same code.

Yes, I think for this rule `L` and `D` have the same form. However, the comment itself should be consistent with others. In other case, we should distinguish floating point data types. That is why I choose to use `L` instead of `D`.

nsjian: If you really want to have a correct comment, you can comment as something like "(L/D)" - though I doubt if it really make any sense for your testing/debugging.

src/hotspot/cpu/aarch64/aarch64_sve.ad
nsjian: I think you can simply fix as what NEON does, like:

+      case Op_VectorLoadShuffle:
+      case Op_VectorRearrange:
+        if (vlen < 4) {
+          return false;
+        }

When we have all codegen ready, I will create another patch to fine-tune the supported vector size for each Op to get best performance.

Wanghuang-Huawei: Yes, I have fixed in this form and tested. I think that full-test will be completed soon.

src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -2411,6 +2411,16 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType
...
+    switch (opcode) {
+      case Op_VectorLoadShuffle:
+      case Op_VectorRearrange:
+        if (vlen < 4) {
+          return false;
+        }
+        break;
+      default:
+        break;
+    }
```
nsjian, line 2423: I would prefer put this inside op_sve_supported.

src/hotspot/cpu/aarch64/aarch64_sve.ad
```diff
@@ -4057,7 +4057,7 @@ instruct rearrangeL(vReg dst, vReg src, vReg shuffle)
             n->bottom_type()->is_vect()->element_basic_type() == T_LONG);
   match(Set dst (VectorRearrange src shuffle));
   ins_cost(SVE_COST);
-  format %{ "sve_tbl $dst, D, $src, $shuffle\t# vector rearrange (D)" %}
+  format %{ "sve_tbl $dst, D, $src, $shuffle\t# vector rearrange (L)" %}
```
nsjian, line 4060: I still think that it make no sense to separate these rules. But anyway, I can do a cleanup later. FYI: Xiaohong and our Intel friends are working hard trying not to increase the instruction selection patterns on masking support design: http://openjdk.java.net/jeps/8261663.

nsjian (approved):
Thanks for the work! Overall looks good to me. Just some nits.

src/hotspot/cpu/aarch64/aarch64.ad
Wanghuang-Huawei: Yes, I will fix that.

PR #80, opened 2021-05-10
8266621: Add masking support for unary/ternary vector intrinsics

XiaohongGong:
Similar with [1], this patch adds the masking support for `unary/ternary` vector intrinsics. It adds the mask information to the original intrinsic methods which are called both by masked and non-masked unary/ternary operations. For non-masked operations, the mask class and value are set to null.

Also change to use the `"binaryMaskedOp"` for vector mask logical operations instead of` "binaryOp"`. So that all the original nary intrinsics and hotspot implementation can be removed.

Note that this patch only contains the Vector API java implementation and the hotspot intrinsics changes. No compiler mid-end and backend implementations are included.

[1] https://bugs.openjdk.java.net/browse/JDK-8264563

---------
### Progress

### Issue
 * [JDK-8266621](https://bugs.openjdk.java.net/browse/JDK-8266621): Add masking support for unary/ternary vector intrinsics

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 242a5e891043d6c6e91db19309367268d0453ffe

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/80/head:pull/80` \
`$ git checkout pull/80`

`$ git checkout pull/80` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/80/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 80`

`$ git pr show -t 80`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/80.diff">https://git.openjdk.java.net/panama-vector/pull/80.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/vmIntrinsics.hpp
+++ b/src/hotspot/share/classfile/vmIntrinsics.hpp
@@ -809,20 +809,20 @@ class methodHandle;
                                                                                                                                                \
   /* Vector API intrinsification support */                                                                                                    \
                                                                                                                                                \
-  do_intrinsic(_VectorUnaryMaskedOp, jdk_internal_vm_vector_VectorSupport, vector_unary_masked_op_name, vector_unary_masked_op_sig, F_S)       \
-   do_signature(vector_unary_masked_op_sig, "(ILjava/lang/Class;Ljava/lang/Class;Ljava/lang/Class;ILjava/lang/Object;Ljava/lang/Object;"       \
-                                             "Ljdk/internal/vm/vector/VectorSupport$UnaryMaskedOperation;)Ljava/lang/Object;")                 \
-   do_name(vector_unary_masked_op_name,     "unaryMaskedOp")                                                                                   \
+  do_intrinsic(_VectorUnaryOp, jdk_internal_vm_vector_VectorSupport, vector_unary_op_name, vector_unary_op_sig, F_S)                           \
+   do_signature(vector_unary_op_sig, "(ILjava/lang/Class;Ljava/lang/Class;Ljava/lang/Class;ILjava/lang/Object;Ljava/lang/Object;"              \
+                                      "Ljdk/internal/vm/vector/VectorSupport$UnaryOperation;)Ljava/lang/Object;")                              \
+   do_name(vector_unary_op_name,     "unaryOp")                                                                                                \
                                                                                                                                                \
-  do_intrinsic(_VectorBinaryMaskedOp, jdk_internal_vm_vector_VectorSupport, vector_binary_masked_op_name, vector_binary_masked_op_sig, F_S)    \
-   do_signature(vector_binary_masked_op_sig, "(ILjava/lang/Class;Ljava/lang/Class;Ljava/lang/Class;ILjava/lang/Object;Ljava/lang/Object;"      \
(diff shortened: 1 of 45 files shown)


PaulSandoz (approved):
Java changes look good.

XiaohongGong:
Thanks for your review @PaulSandoz !

XiaohongGong:
@iwanowww , could you please take a look at this PR? Thanks so much!

XiaohongGong:
Since all the `naryOp ` are totally replaced with `naryMaskedOp` (e.g. `binaryOp -> binaryMaskedOp`), does it make sense that we still use the name `naryOp` for both masked and non-masked operations? @PaulSandoz which do you think is better `binaryOp` or `binaryMaskedOp` ?

PaulSandoz:
I think what you have is fine, lets get the code in the branch and iterate. It's easy to revisit naming decisions like this later.

XiaohongGong:
> I think what you have is fine, lets get the code in the branch and iterate. It's easy to revisit naming decisions like this later.

OK, agree! Thanks so much for your advice!

iwanowww (commented):
Overall, looks good.

2 comments:
* On naming: while browsing the patch and seeing `Masked` I constantly think about non-Masked counterpart. What about reusing `unaryOp`/`binaryOp`/`ternaryOp` names for the merged intrinsic?  

* Can you remind me what happens for SVML stubs when `mask != NULL`? Does the JVM reject the intrinsification? I believe the stubs don't support masked operation mode.

XiaohongGong:
> Overall, looks good.

Thanks for your review!

> 
> 2 comments:
> 
> * On naming: while browsing the patch and seeing `Masked` I constantly think about non-Masked counterpart. What about reusing `unaryOp`/`binaryOp`/`ternaryOp` names for the merged intrinsic?

Yeah, totally agree. I also pointed out the naming issue above (see: https://github.com/openjdk/panama-vector/pull/80#issuecomment-840358060). And I agree with what @PaulSandoz's comment about it: "lets get the code in the branch and iterate. It's easy to revisit naming decisions like this later.". Does this make sense to you?

> Can you remind me what happens for SVML stubs when mask != NULL? Does the JVM reject the intrinsification? I believe the stubs don't support masked operation mode.

It will generate the vector blend mode like before. Currently the compiler only checks whether the predicated mode is supported for normal operations `(sopc != 0)`. For SVML ops, the `use_predicate` is false which means it generates `"VectorBlend"` pattern like before.

XiaohongGong:
Hi @iwanowww , do you have any other comments for this PR? Can we merge this PR first and revisit the naming issue after then?

XiaohongGong:
@PaulSandoz  @iwanowww , I'v rename all the masked op back to the original names. Could you please help to look at again? Waiting for your feedback, Thanks so much!

iwanowww (approved):
Looks good.

PaulSandoz:
> @PaulSandoz @iwanowww , I'v rename all the masked op back to the original names. Could you please help to look at again? Waiting for your feedback, Thanks so much!

Looks good. Integrate it!

XiaohongGong:
Thanks for the review @PaulSandoz @iwanowww !

PR #81, opened 2021-05-12
8266720: Wrong implementation in LibraryCallKit::inline_vector_shuffle_iota

Wanghuang-Huawei:
Because JDK-8266317 has not been merged into jdk/jdk. So I fix this bug here.
* This comparsion should be a unsigned cmp
*  `x86` is not wrong because `x86` does not have 1024-bits vl here. 
  ```c++

    ConINode* pred_node = (ConINode*)gvn().makecon(TypeInt::make(BoolTest::ge)); // should BoolTest::ugt
    Node * lane_cnt  = gvn().makecon(TypeInt::make(num_elem));
    Node * bcast_lane_cnt = gvn().transform(VectorNode::scalar2vector(lane_cnt, num_elem, type_bt));
    // should BoolTest::ugt
    Node* mask = gvn().transform(new VectorMaskCmpNode(BoolTest::ge, bcast_lane_cnt, res, pred_node, vt));

```

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Error
&nbsp;⚠️ Executable files are not allowed (file: src/jdk.incubator.vector/share/classes/jdk/incubator/vector/gen-src.sh)

### Integration blockers
&nbsp;⚠️ Executable files are not allowed (file: src/jdk.incubator.vector/share/classes/jdk/incubator/vector/gen-src.sh) (failed with the updated jcheck configuration)
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Issue
 * [JDK-8266720](https://bugs.openjdk.org/browse/JDK-8266720): Wrong implementation in LibraryCallKit::inline_vector_shuffle_iota (**Bug** - P2)

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-vector.git pull/81/head:pull/81` \
`$ git checkout pull/81`

`$ git checkout pull/81` \
`$ git pull https://git.openjdk.org/panama-vector.git pull/81/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 81`

`$ git pr show -t 81`

</details>

<a href="https://git.openjdk.org/panama-vector/pull/81.diff">https://git.openjdk.org/panama-vector/pull/81.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-vector/pull/81#issuecomment-839628736)

### CODE REVIEW

File: test/hotspot/jtreg/compiler/vectorapi/TestVectorShuffleIota1024.java [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+/*
+ * @test
+ * @bug 8266720
+ * @modules jdk.incubator.vector
+ * @run testng/othervm compiler.vectorapi.TestVectorShuffleIotaByte1024
```
PaulSandoz, line 38: Perhaps the test can be annotated, declaring it should only execute on ARM/SVE platforms. See the use of the `@requires` clause used in other JDK tests.

Wanghuang-Huawei: Thank you for you review. I think this test is used for any arch which has `ByteVector.SPECIES_MAX == 1024`.

PaulSandoz: But we know which arches don't support, x86, PPC etc.

I am unsure why existing shuffle tests do not catch this problem. In fact i would prefer we focus on that if we can rather than adding a specific test. Would you mind looking to see if see if we can expand on the existing `shuffleTest`?

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -453,10 +453,10 @@ bool LibraryCallKit::inline_vector_shuffle_iota() {
     // Wrap the indices greater than lane count.
     res = gvn().transform(VectorNode::make(Op_AndI, res, bcast_mod, num_elem, elem_bt));
   } else {
-    ConINode* pred_node = (ConINode*)gvn().makecon(TypeInt::make(BoolTest::ge));
+    ConINode* pred_node = (ConINode*)gvn().makecon(TypeInt::make(BoolTest::ugt));
```
sviswa7, line 456: Unsigned comparison adds overhead and is not supported on all architectures.

Wanghuang-Huawei: 1 


After changing notes with @XiaohongGong , I think we can also fix like this: 
```java
    ConINode* pred_node = (ConINode*)gvn().makecon(TypeInt::make(BoolTest::ge));
    Node * lane_cnt_tmp  = gvn().makecon(TypeInt::make(num_elem - 1));
    Node * bcast_lane_cnt = gvn().transform(VectorNode::scalar2vector(lane_cnt_tmp, num_elem, type_bt));
    Node* mask = gvn().transform(new VectorMaskCmpNode(BoolTest::ge, bcast_lane_cnt, res, pred_node, vt));

    // Make the indices greater than lane count as -ve values. This matches the java side implementation.
    res = gvn().transform(VectorNode::make(Op_AndI, res, bcast_mod, num_elem, elem_bt));
    Node * lane_cnt  = gvn().makecon(TypeInt::make(num_elem)); // Add a mov & bcast here
    Node * bcast_lane_cnt = gvn().transform(VectorNode::scalar2vector(lane_cnt, num_elem, type_bt));
    Node * biased_val = gvn().transform(VectorNode::make(Op_SubI, res, bcast_lane_cnt, num_elem, elem_bt));
    res = gvn().transform(new VectorBlendNode(biased_val, res, mask));

```

Wanghuang-Huawei: > Unsigned comparison adds overhead and is not supported on all architectures.

However, if we don't use ugt. We will encounter problem if length > 1024  in future.  Changing `< num_elem` to `<= num_elem` is just a solution to `1024` itself. If `num_elem > 1024`, it will be invalid.

XiaohongGong: Currently making it work well for <= 1024-bits makes sense to me. We can revisit this issue after the API issues for vector length > 1024-bits are fixed in future.

nsjian: > Currently making it work well for <= 1024-bits makes sense to me. We can revisit this issue after the API issues for vector length > 1024-bits are fixed in future.

If so, we need at least some comments or even length check to not inline for unsupported vector length?

XiaohongGong: Agree with @nsjian ! Thanks!

Wanghuang-Huawei: If we use `ge 1024` ,  we will add two more nodes. It is a extra cost here?

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -454,11 +454,23 @@ bool LibraryCallKit::inline_vector_shuffle_iota() {
...
     Node * lane_cnt  = gvn().makecon(TypeInt::make(num_elem));
     Node * bcast_lane_cnt = gvn().transform(VectorNode::scalar2vector(lane_cnt, num_elem, type_bt));
-    Node* mask = gvn().transform(new VectorMaskCmpNode(BoolTest::ugt, bcast_lane_cnt, res, pred_node, vt));
 
+    if (Matcher::supports_unsigned_vector_comparison(num_elem, elem_bt)) {
+      pred_node = (ConINode*)gvn().makecon(TypeInt::make(BoolTest::ugt));
+      mask = gvn().transform(new VectorMaskCmpNode(BoolTest::ugt, bcast_lane_cnt, res, pred_node, vt));
+    } else {
+      // Currently it works well for vector_length <= 1024-bits.
+      // for vector_length > 1024, we don't support now
```
XiaohongGong, line 467: Does it need any vector length check or block for vector length >1024?

Wanghuang-Huawei: thank you for your reviw. It is a problem here. I will do that in next commit.

test/hotspot/jtreg/compiler/vectorapi/TestVectorShuffleIota1024.java
Wanghuang-Huawei: * Q : Why existing shuffle tests do not catch this problem? 
* A: Because we need `vector_length >= 1024`. However, in x86 we don't have this env because the longest register of x86 is 512 in AVX512.

src/hotspot/share/opto/vectorIntrinsics.cpp
XiaohongGong: As discussed with @nsjian , can we currently make it unsupported as well for 1024-bits? It doesn't need to change any IR if 1024-bits is not supported. We can revisit this issue in future for vector length >= 1024-bits

PR #90, opened 2021-06-07
8268154: Add masking support for vector load intrinsics

XiaohongGong:
This patch adds the predicate support for masked load vector operation, including the Vector API java implementation and hotspot intrinsics changes. The main changes contain:
  - Adding a new hotspot intrinsic method (`loadMasked`), which can transform the vector mask class and value to hotspot. It is called by the masked `"fromArray"` API in JAVA level. The compiler will generate the masked vector load node if the current platform supports the predicate feature and backend has implemented it. Otherwise, the vector blend pattern will be generated like before.
  - Adding the similar masked support for load/store boolean/char array.

Note that the intrinsification when there is the array range bailout is not implemented in this patch. We are considering making it vectorized with predicate feature for SVE/AVX-512 as well, and will create a separate patch for it in future.

Also note that this patch doesn't contain any backend changes.

---------
### Progress

### Issue
 * [JDK-8268154](https://bugs.openjdk.java.net/browse/JDK-8268154): Add masking support for vector load intrinsics

### Reviewers
 * [Jatin Bhateja](https://openjdk.java.net/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to 0b77c8e33df2676d6497e013e3e29769747a6283
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 106585f62fd380d09c7c8d23a7f6d54f44f698a9

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/90/head:pull/90` \
`$ git checkout pull/90`

`$ git checkout pull/90` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/90/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 90`

`$ git pr show -t 90`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/90.diff">https://git.openjdk.java.net/panama-vector/pull/90.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -186,7 +186,7 @@ bool LibraryCallKit::arch_supports_vector(int sopc, int num_elem, BasicType type
   if ((mask_use_type & VecMaskUsePred) != 0) {
     if (!Matcher::has_predicated_vectors() ||
         !Matcher::match_rule_supported_vector_masked(sopc, num_elem, type)) {
-      #ifndef PRODUCT
+    #ifndef PRODUCT
       if (C->print_intrinsics()) {
         tty->print_cr("Rejected vector mask predicate using (%s,%s,%d) because architecture does not support it",
                       NodeClassNames[sopc], type2name(type), num_elem);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
@@ -4195,7 +4195,6 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
         if (m.allTrue()) {
             intoCharArray(a, offset);
         } else {
-            // FIXME: optimize
             $Type$Species vsp = vspecies();
             checkMaskFromIndexSize(offset, vsp, m, 1, a.length);
             intoCharArray0(a, offset, m);
@@ -4358,7 +4357,6 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
         if (m.allTrue()) {
             intoBooleanArray(a, offset);
         } else {
-            // FIXME: optimize
             $Type$Species vsp = vspecies();
             checkMaskFromIndexSize(offset, vsp, m, 1, a.length);
             intoBooleanArray0(a, offset, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
@@ -3403,7 +3403,6 @@ public abstract class ByteVector extends AbstractVector<Byte> {
         if (m.allTrue()) {
             intoBooleanArray(a, offset);
         } else {
-            // FIXME: optimize
             ByteSpecies vsp = vspecies();
             checkMaskFromIndexSize(offset, vsp, m, 1, a.length);
             intoBooleanArray0(a, offset, m);
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ShortVector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ShortVector.java
@@ -3394,7 +3394,6 @@ public abstract class ShortVector extends AbstractVector<Short> {
         if (m.allTrue()) {
             intoCharArray(a, offset);
         } else {
-            // FIXME: optimize
             ShortSpecies vsp = vspecies();
             checkMaskFromIndexSize(offset, vsp, m, 1, a.length);
             intoCharArray0(a, offset, m);


XiaohongGong:
Hi @PaulSandoz , @sviswa7 , could you please take a look at the java side of this PR? Thanks so much!

Hi @iwanowww , could you please take a look at the compiler side of this PR? Thanks so much! Regarding to your feedback!

Best Regards,
Xiaohong

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -948,18 +947,32 @@ bool LibraryCallKit::inline_vector_mem_masked_operation(bool is_store) {
...
-
   BasicType elem_bt = elem_type->basic_type();
   int num_elem = vlen->get_con();
+
   int sopc = is_store ? Op_StoreVectorMasked : Op_LoadVectorMasked;
-  if (!arch_supports_vector(sopc, num_elem, elem_bt, VecMaskUseLoad) ||
-      !Matcher::match_rule_supported_vector_masked(sopc, num_elem, elem_bt)) {
+  bool use_predicate = Matcher::has_predicated_vectors() &&
+                       arch_supports_vector(sopc, num_elem, elem_bt, VecMaskUseLoad) &&
+                       Matcher::match_rule_supported_vector_masked(sopc, num_elem, elem_bt);
```
jatin-bhateja, line 956: Can we push both  Matcher::has_predicated_vectors() && Matcher::match_rule_supported_vector_masked() into arch_supports_vector and either add an explicit bool argument to it or a new enum value to be passed i.e. VecMaskUseLoad | VecPredOper  under which these checks are enforced.

jatin-bhateja (commented):
Currently there following assertion check in both LoadVectorMaskedNode and StoreVectorMaskedNode  

`assert(mask->bottom_type()->is_vectmask(), "sanity");`

This should be relaxed since mask argument will be VectorUnboxNode until macro expansion.

I added following static routine in VectorNode class to get over it in my local sandbox, and adjusted the assertion to use it. If suitable maybe we can integrate it in this patch.

```
bool VectorNode::is_vector_mask(Node* n) {
  if (n->bottom_type()->is_vectmask()) {
    return true;
  }
  if (n->Opcode() == Op_VectorUnbox) {
    const TypeInstPtr* vector_klass_from = n->as_VectorUnbox()->obj()->bottom_type()->isa_instptr();
    ciKlass* vbox_klass_from = vector_klass_from->const_oop()->as_instance()->java_lang_Class_klass();
    return vbox_klass_from->is_subclass_of(ciEnv::current()->vector_VectorMask_klass());
  }
  return false;
}
```

jatin-bhateja:
Over all patch looks good to me.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -4565,6 +4563,23 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
...
+    final
+    <M extends VectorMask<$Boxtype$>>
+    $abstractvectortype$ fromArray0Template(Class<M> maskClass, $type$[] a, int offset, M m) {
+        m.check(species());
+        $Type$Species vsp = vspecies();
+        return VectorSupport.loadMasked(
+            vsp.vectorType(), maskClass, vsp.elementType(), vsp.laneCount(),
+            a, arrayAddress(a, offset), m,
+            a, offset, vsp,
+            (arr, off, s, vm) -> s.ldOp(arr, off, (AbstractMask<$Boxtype$>) vm,
```
PaulSandoz, line 4579: We can avoid the cast to `AbstractMask<$Boxtype$>` if we update the signature of the species `ldOp` to be e.g on `IntVector`:
```java
        <M> IntVector ldOp(M memory, int offset,
                                      VectorMask<Integer> m,
                                      FLdOp<M> f) {
            return dummyVector().ldOp(memory, offset, m, f);
        }
```

In general i think we need to review the generic signatures here and in `VectorSupport`, as they are a bit inconsistent. Something to consider later.

PaulSandoz (approved):
Java changes look good. No need for me to re-review to resolve the comment i made.

src/hotspot/share/opto/vectorIntrinsics.cpp
XiaohongGong: Good idea! Will do this later. Thanks so much!

XiaohongGong:
> Currently there following assertion check in both LoadVectorMaskedNode and StoreVectorMaskedNode
> 
> `assert(mask->bottom_type()->is_vectmask(), "sanity");`
> 
> This should be relaxed since mask argument will be VectorUnboxNode until macro expansion.
> 
> I added following static routine in VectorNode class to get over it in my local sandbox, and adjusted the assertion to use it. If suitable maybe we can integrate it in this patch.
> 
> ```
> bool VectorNode::is_vector_mask(Node* n) {
>   if (n->bottom_type()->is_vectmask()) {
>     return true;
>   }
>   if (n->Opcode() == Op_VectorUnbox) {
>     const TypeInstPtr* vector_klass_from = n->as_VectorUnbox()->obj()->bottom_type()->isa_instptr();
>     ciKlass* vbox_klass_from = vector_klass_from->const_oop()->as_instance()->java_lang_Class_klass();
>     return vbox_klass_from->is_subclass_of(ciEnv::current()->vector_VectorMask_klass());
>   }
>   return false;
> }
> ```

Thanks so for your suggestion @jatin-bhateja ! This looks ok to me. So that I can also revert the type changes to VectorBox/VectorUnbox in PR https://github.com/openjdk/panama-vector/pull/78.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
XiaohongGong: Thanks for looking at this PR @PaulSandoz ! This looks ok to me and will change it later.

XiaohongGong:
> > Currently there following assertion check in both LoadVectorMaskedNode and StoreVectorMaskedNode
> > `assert(mask->bottom_type()->is_vectmask(), "sanity");`
> > This should be relaxed since mask argument will be VectorUnboxNode until macro expansion.
> > I added following static routine in VectorNode class to get over it in my local sandbox, and adjusted the assertion to use it. If suitable maybe we can integrate it in this patch.
> > ```
> > bool VectorNode::is_vector_mask(Node* n) {
> >   if (n->bottom_type()->is_vectmask()) {
> >     return true;
> >   }
> >   if (n->Opcode() == Op_VectorUnbox) {
> >     const TypeInstPtr* vector_klass_from = n->as_VectorUnbox()->obj()->bottom_type()->isa_instptr();
> >     ciKlass* vbox_klass_from = vector_klass_from->const_oop()->as_instance()->java_lang_Class_klass();
> >     return vbox_klass_from->is_subclass_of(ciEnv::current()->vector_VectorMask_klass());
> >   }
> >   return false;
> > }
> > ```
> 
> Thanks so for your suggestion @jatin-bhateja ! This looks ok to me. So that I can also revert the type changes to VectorBox/VectorUnbox in PR #78.

Regarding to merge it, do you think it's better to fix it in https://github.com/openjdk/panama-vector/pull/78 instead of this PR? One reason is that there are other mask IRs that also contains the same assertion, we can update them together. Another one is that I can revert the type changes for `VectorBox/VectorUnbox` and have a test together. Thanks!

XiaohongGong:
Hi @PaulSandoz @jatin-bhateja , the latest commit have addressed your comments. Thanks!

jatin-bhateja:
Thanks!, changes looks good to me.

jatin-bhateja (approved):

XiaohongGong:
Thanks for the review @jatin-bhateja @PaulSandoz !

jatin-bhateja:
Hi @XiaohongGong , can you kindly push these common IR changes

XiaohongGong:
> Hi @XiaohongGong , can you kindly push these common IR changes

Sure. I was intending to wait for the feedback from @iwanowww . Maybe it's better to integrate these changes first to make sure the masking feature support is not blocked. Thanks so much for the advice and review!

@iwanowww , I'd like to integrate this PR first. If there are any comments or feedback, please feel free to contact me and we can revisit them with a followed-up patch. Thanks so much!

PR #96, opened 2021-06-25
8269343: Masked vector arithmetic intrinsics failed to be inlined randomly

XiaohongGong:
The masked `"lanewise/reduceLanes"` failed to be intrinsified randomly. The main reason is the hotspot c2 compiler cannot
unbox the mask value due to the expected vector box type being mismatched with the class of the mask value. The vector box type is the` "mask class"` which is specified in Java level and saved in the intrinsic argument list. It is the class of the instance
of the vector mask (i.e. `"IntMaxMask.class"`), while the type of mask value is the super class (i.e. `"jdk.incubator.vector.VectorMask<Integer>"`).

Casting the mask value to the same subclass before calling the intrinsics can solve this issue.

---------
### Progress

### Issue
 * [JDK-8269343](https://bugs.openjdk.java.net/browse/JDK-8269343): Masked vector arithmetic intrinsics failed to be inlined randomly

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/96/head:pull/96` \
`$ git checkout pull/96`

`$ git checkout pull/96` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/96/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 96`

`$ git pr show -t 96`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/96.diff">https://git.openjdk.java.net/panama-vector/pull/96.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
@@ -276,7 +276,7 @@ final class Byte128Vector extends ByteVector {
     @Override
     @ForceInline
     public Byte128Vector lanewise(Unary op, VectorMask<Byte> m) {
-        return (Byte128Vector) super.lanewiseTemplate(op, Byte128Mask.class, m);  // specialize
+        return (Byte128Vector) super.lanewiseTemplate(op, Byte128Mask.class, (Byte128Mask) m);  // specialize
     }
 
     @Override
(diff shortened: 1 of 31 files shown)


XiaohongGong:
@PaulSandoz , could you please take a look at this PR? Thanks so much!

PaulSandoz (approved):
Looks good.

I am wondering why its not needed on the vector arguments, is that because there is a cast to `{T}Vector` for vector arguments in templates is sufficient in those cases? Or does the `that.check(this);` help as well?

XiaohongGong:
> Looks good.
> 
Thanks for your review @PaulSandoz !

> I am wondering why its not needed on the vector arguments, is that because there is a cast to `{T}Vector` for vector arguments in templates is sufficient in those cases? Or does the `that.check(this);` help as well?
>
I think one reason is the vector class is transfered by calling `this.getClass()` when calling the intrinsic method. And  `that.check(this);` also helps since there is the vector class equalling check inside this method.

I also tried to get the mask class by calling `m.getClass()` like the vector class, but the compiler cannot guarantee the class is a constant which could also make the intrinsic fail to inline. As a reference to the existing `blend` API, it also explicitly set the mask class to the instance of the subclass which is sure to be a constant.

PR #99, opened 2021-07-22
8270349: Initial X86 backend support for optimizing masking operations on AVX512 targets.

jatin-bhateja:
Intel targets supporting AVX512 feature offer predicated vector instructions. These are vector operations on selected vector lanes under the influence of opmask register. For non-AVX512 targets, masked vector operations are supported using an explicit vector blend operation after main vector operation which does the needed selection. 

This patch adds initial X86 backed support for predicated vector operations. 

Following is performance data for existing VectorAPI JMH benchmarks with the patch:
Test System:  Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz (Icelake server 40C 2S)

Benchmark | SIZE | Baseline (ops/ms) | WithOpts (ops/ms) | Gain
-- | -- | -- | -- | --
Int512Vector.ABSMasked | 1024 | 10132.664 | 10394.942 | 1.025884407
Int512Vector.ADDMasked | 1024 | 7785.805 | 8980.133 | 1.153398139
Int512Vector.ADDMaskedLanes | 1024 | 5809.455 | 6350.628 | 1.093153833
Int512Vector.ANDMasked | 1024 | 7776.729 | 8965.988 | 1.152925349
Int512Vector.ANDMaskedLanes | 1024 | 6717.202 | 7426.217 | 1.105552133
Int512Vector.AND_NOTMasked | 1024 | 7688.835 | 8988.659 | 1.169053439
Int512Vector.ASHRMasked | 1024 | 6808.185 | 7883.755 | 1.1579819
Int512Vector.ASHRMaskedShift | 1024 | 9523.164 | 12166.72 | 1.277592195
Int512Vector.BITWISE_BLENDMasked | 1024 | 5919.647 | 6864.988 | 1.159695502
Int512Vector.DIVMasked | 1024 | 237.174 | 236.014 | 0.995109076
Int512Vector.FIRST_NONZEROMasked | 1024 | 5387.315 | 7890.42 | 1.464629412
Int512Vector.LSHLMasked | 1024 | 6806.898 | 7881.315 | 1.157842383
Int512Vector.LSHLMaskedShift | 1024 | 9552.257 | 12153.769 | 1.272345269
Int512Vector.LSHRMasked | 1024 | 6776.605 | 7897.786 | 1.165448776
Int512Vector.LSHRMaskedShift | 1024 | 9500.087 | 12134.962 | 1.277352723
Int512Vector.MAXMaskedLanes | 1024 | 6993.149 | 7580.399 | 1.083975045
Int512Vector.MINMaskedLanes | 1024 | 6925.363 | 7450.814 | 1.075873424
Int512Vector.MULMasked | 1024 | 7732.753 | 8956.02 | 1.158192949
Int512Vector.MULMaskedLanes | 1024 | 4066.384 | 4152.375 | 1.021146798
Int512Vector.NEGMasked | 1024 | 8760.797 | 9255.063 | 1.056417926
Int512Vector.NOTMasked | 1024 | 8981.123 | 9229.573 | 1.027663578
Int512Vector.ORMasked | 1024 | 7786.787 | 8967.057 | 1.151573428
Int512Vector.ORMaskedLanes | 1024 | 6694.36 | 7450.106 | 1.112892943
Int512Vector.SUBMasked | 1024 | 7782.939 | 9001.692 | 1.156592901
Int512Vector.XORMasked | 1024 | 7785.031 | 9070.342 | 1.165100306
Int512Vector.XORMaskedLanes | 1024 | 6700.689 | 7454.73 | 1.112531861
Int512Vector.ZOMOMasked | 1024 | 6982.297 | 8313.51 | 1.190655453
Int512Vector.gatherMasked | 1024 | 361.497 | 1494.876 | 4.135237637
Int512Vector.scatterMasked | 1024 | 490.05 | 3120.425 | 6.367564534
Int512Vector.sliceMasked | 1024 | 1436.248 | 1597.805 | 1.112485448
Int512Vector.unsliceMasked | 1024 | 296.721 | 346.434 | 1.167541226
Float512Vector.ADDMasked | 1024 | 7645.873 | 9123.386 | 1.193243205
Float512Vector.ADDMaskedLanes | 1024 | 2404.371 | 2529.284 | 1.051952465
Float512Vector.DIVMasked | 1024 | 5134.602 | 5129.085 | 0.998925525
Float512Vector.FIRST_NONZEROMasked | 1024 | 5040.567 | 7078.828 | 1.404371373
Float512Vector.FMAMasked | 1024 | 5996.419 | 6902.626 | 1.151124696
Float512Vector.MAXMaskedLanes | 1024 | 1681.249 | 1727.444 | 1.027476596
Float512Vector.MINMaskedLanes | 1024 | 1610.115 | 1667.143 | 1.035418588
Float512Vector.MULMasked | 1024 | 7812.317 | 9054.137 | 1.158956683
Float512Vector.MULMaskedLanes | 1024 | 2406.81 | 2514.018 | 1.044543608
Float512Vector.NEGMasked | 1024 | 8248.933 | 9834.607 | 1.192227771
Float512Vector.SQRTMasked | 1024 | 4278.046 | 4281.009 | 1.000692606
Float512Vector.SUBMasked | 1024 | 7697.582 | 9044.305 | 1.174954031
Float512Vector.gatherMasked | 1024 | 428.428 | 1491.441 | 3.48119404
Float512Vector.scatterMasked | 1024 | 416.169 | 3216.628 | 7.729138883
Float512Vector.sliceMasked | 1024 | 1431.07 | 1609.12 | 1.124417394
Float512Vector.unsliceMasked | 1024 | 292.513 | 331.366 | 1.132824866

PS: Above data shows the performance gains for two vector species Int512, Float512.  In general for all the species we see 1.2-2.x gains on various masking operation supported uptill now.
New matcher routine `Matcher::match_rule_supported_vector_masked`   lists making operations supported by this patch.

---------
### Progress

### Issue
 * [JDK-8270349](https://bugs.openjdk.java.net/browse/JDK-8270349): Initial X86 backend support for optimizing masking operations on AVX512 targets.

### Reviewers
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to b001f939c60f14a832830d16445d16f58d7ad865

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/99/head:pull/99` \
`$ git checkout pull/99`

`$ git checkout pull/99` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/99/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 99`

`$ git pr show -t 99`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/99.diff">https://git.openjdk.java.net/panama-vector/pull/99.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -3440,7 +3440,8 @@ instruct sqrtD_reg(regD dst) %{
 
 // ---------------------------------------- VectorReinterpret ------------------------------------
 instruct reinterpret_mask(kReg dst) %{
-  predicate(UseAVX > 2 && Matcher::vector_length(n) == Matcher::vector_length(n->in(1))); // dst == src
+  predicate(n->bottom_type()->isa_vectmask() &&
+            Matcher::vector_length(n) == Matcher::vector_length(n->in(1))); // dst == src
   match(Set dst (VectorReinterpret dst));
   ins_cost(125);
   format %{ "vector_reinterpret $dst\t!" %}
@@ -3452,6 +3453,7 @@ instruct reinterpret_mask(kReg dst) %{
 
 instruct reinterpret_mask_W2B(kReg dst, kReg src, vec xtmp) %{
   predicate(UseAVX > 2 && Matcher::vector_length(n) != Matcher::vector_length(n->in(1)) &&
+            n->bottom_type()->isa_vectmask() &&
             n->as_VectorReinterpret()->src_elem_type() == T_SHORT &&
             n->as_VectorReinterpret()->dst_elem_type() == T_BYTE); // dst == src
   match(Set dst (VectorReinterpret src));
@@ -3470,6 +3472,7 @@ instruct reinterpret_mask_W2B(kReg dst, kReg src, vec xtmp) %{
 
 instruct reinterpret_mask_D2B(kReg dst, kReg src, vec xtmp) %{
   predicate(UseAVX > 2 && Matcher::vector_length(n) != Matcher::vector_length(n->in(1)) &&
+            n->bottom_type()->isa_vectmask() &&
             (n->as_VectorReinterpret()->src_elem_type() == T_INT ||
             n->as_VectorReinterpret()->src_elem_type() == T_FLOAT) &&
             n->as_VectorReinterpret()->dst_elem_type() == T_BYTE); // dst == src
@@ -3489,6 +3492,7 @@ instruct reinterpret_mask_D2B(kReg dst, kReg src, vec xtmp) %{
 
 instruct reinterpret_mask_Q2B(kReg dst, kReg src, vec xtmp) %{
   predicate(UseAVX > 2 && Matcher::vector_length(n) != Matcher::vector_length(n->in(1)) &&
+            n->bottom_type()->isa_vectmask() &&
             (n->as_VectorReinterpret()->src_elem_type() == T_LONG ||
             n->as_VectorReinterpret()->src_elem_type() == T_DOUBLE) &&
             n->as_VectorReinterpret()->dst_elem_type() == T_BYTE); // dst == src
@@ -3507,7 +3511,8 @@ instruct reinterpret_mask_Q2B(kReg dst, kReg src, vec xtmp) %{
 %}
 
 instruct reinterpret(vec dst) %{
-  predicate(Matcher::vector_length_in_bytes(n) == Matcher::vector_length_in_bytes(n->in(1))); // dst == src
+  predicate(!n->bottom_type()->isa_vectmask() &&
+            Matcher::vector_length_in_bytes(n) == Matcher::vector_length_in_bytes(n->in(1))); // dst == src
   match(Set dst (VectorReinterpret dst));
   ins_cost(125);
   format %{ "vector_reinterpret $dst\t!" %}
@@ -3542,6 +3547,7 @@ instruct reinterpret_expand(vec dst, vec src, rRegP scratch) %{
 
 instruct vreinterpret_expand4(legVec dst, vec src, rRegP scratch) %{
   predicate(UseAVX > 0 &&
+            !n->bottom_type()->isa_vectmask() &&
             (Matcher::vector_length_in_bytes(n->in(1)) == 4) && // src
             (Matcher::vector_length_in_bytes(n->in(1)) < Matcher::vector_length_in_bytes(n))); // src < dst
   match(Set dst (VectorReinterpret src));
@@ -3557,6 +3563,7 @@ instruct vreinterpret_expand4(legVec dst, vec src, rRegP scratch) %{
 
 instruct vreinterpret_expand(legVec dst, vec src) %{
   predicate(UseAVX > 0 &&
+            !n->bottom_type()->isa_vectmask() &&
             (Matcher::vector_length_in_bytes(n->in(1)) > 4) && // src
             (Matcher::vector_length_in_bytes(n->in(1)) < Matcher::vector_length_in_bytes(n))); // src < dst
   match(Set dst (VectorReinterpret src));
@@ -3574,7 +3581,8 @@ instruct vreinterpret_expand(legVec dst, vec src) %{
 %}
 
 instruct reinterpret_shrink(vec dst, legVec src) %{
-  predicate(Matcher::vector_length_in_bytes(n->in(1)) > Matcher::vector_length_in_bytes(n)); // src > dst
+  predicate(!n->bottom_type()->isa_vectmask() &&
+            Matcher::vector_length_in_bytes(n->in(1)) > Matcher::vector_length_in_bytes(n)); // src > dst
   match(Set dst (VectorReinterpret src));
   ins_cost(125);
   format %{ "vector_reinterpret_shrink $dst,$src\t!" %}
@@ -3830,7 +3838,7 @@ instruct evgather_masked(vec dst, memory mem, vec idx, kReg mask, kReg ktmp, rRe
     int vlen_enc = vector_length_encoding(this);
     BasicType elem_bt = Matcher::vector_element_basic_type(this);
     assert(!is_subword_type(elem_bt), "sanity"); // T_INT, T_LONG, T_FLOAT, T_DOUBLE
-    // Note: Since gather instruction partially updates the opmask register used 
+    // Note: Since gather instruction partially updates the opmask register used
     // for predication hense moving mask operand to a temporary.
     __ kmovwl($ktmp$$KRegister, $mask$$KRegister);
     __ vpxor($dst$$XMMRegister, $dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
@@ -3872,7 +3880,7 @@ instruct scatter_masked(memory mem, vec src, vec idx, kReg mask, kReg ktmp, rReg
     BasicType elem_bt = Matcher::vector_element_basic_type(this, $src);
     assert(Matcher::vector_length_in_bytes(this, $src) >= 16, "sanity");
     assert(!is_subword_type(elem_bt), "sanity"); // T_INT, T_LONG, T_FLOAT, T_DOUBLE
-    // Note: Since scatter instruction partially updates the opmask register used 
+    // Note: Since scatter instruction partially updates the opmask register used
     // for predication hense moving mask operand to a temporary.
     __ kmovwl($ktmp$$KRegister, $mask$$KRegister);
     __ lea($tmp$$Register, $mem$$Address);
@@ -7252,7 +7260,7 @@ instruct vcmpu32(legVec dst, legVec src1, legVec src2, immI8 cond, legVec vtmp1,
 
 instruct vcmpu64(vec dst, vec src1, vec src2, immI8 cond, rRegP scratch, kReg ktmp) %{
   predicate(UseAVX > 2 &&
-            (VM_Version::supports_avx512vl() ||
+            (!VM_Version::supports_avx512vl() &&
              Matcher::vector_length_in_bytes(n->in(1)->in(1)) == 64) && // src1
              is_integral_type(Matcher::vector_element_basic_type(n->in(1)->in(1)))); // src1
   match(Set dst (VectorMaskCmp (Binary src1 src2) cond));
--- a/src/hotspot/share/opto/vectornode.hpp
+++ b/src/hotspot/share/opto/vectornode.hpp
@@ -1442,6 +1442,7 @@ class VectorReinterpretNode : public VectorNode {
  public:
   VectorReinterpretNode(Node* in, const TypeVect* src_vt, const TypeVect* dst_vt)
       : VectorNode(in, dst_vt), _src_vt(src_vt) {
+     assert(!dst_vt->isa_vectmask() && !src_vt->isa_vectmask(), "");
      _src_bt = src_vt->element_basic_type();
      _dst_bt = dst_vt->element_basic_type();
      init_class_id(Class_VectorReinterpret);
@@ -1450,6 +1451,8 @@ class VectorReinterpretNode : public VectorNode {
   VectorReinterpretNode(Node* in, BasicType src_bt, const TypeVect* src_vt,
                         BasicType dst_bt, const TypeVect* dst_vt)
       : VectorNode(in, dst_vt), _src_vt(src_vt) {
+     assert(dst_vt->isa_vectmask() && src_vt->isa_vectmask() || type2aelembytes(src_bt) >= type2aelembytes(dst_bt),
+            "unsupported mask widening reinterpretation");
      _src_bt = src_bt;
      _dst_bt = dst_bt;
      init_class_id(Class_VectorReinterpret);


src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -432,7 +432,7 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
       return false;
      }
   } else {
-    const TypeVect* vt = TypeVect::make(elem_bt, num_elem);
+    const TypeVect* vt = TypeVect::make(elem_bt, num_elem, is_vector_mask(vbox_klass));
```
XiaohongGong, line 435: Are there any other nary mask operations besides the `"and/xor/or" `?

src/hotspot/share/opto/vectornode.cpp
```diff
@@ -428,17 +432,17 @@ VectorNode* VectorNode::make_mask_node(int vopc, Node* n1, Node* n2, uint vlen,
   const TypeVect* vmask_type = TypeVect::makemask(bt, vlen);
   switch (vopc) {
     case Op_AndV:
-      if (Matcher::match_rule_supported_vector(Op_AndVMask, vlen, bt)) {
+      if (Matcher::match_rule_supported_vector_masked(Op_AndVMask, vlen, bt)) {
```
XiaohongGong, line 435: I intended to use `match_rule_supported_vector_masked` specially for the vector operations that are controlled by a mask. Do you mean we'd better also extend it to the mask operations?

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -1299,7 +1297,7 @@ class VectorMaskCmpNode : public VectorNode {
   BoolTest::mask _predicate;
 
  protected:
-  uint size_of() const { return sizeof(*this); }
+  virtual  uint  size_of() const { return sizeof(VectorMaskCmpNode); }
```
XiaohongGong, line 1300: One more space between `"uint   size_of"`?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -1835,12 +1838,104 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType
...
+        return false; // Implementation limitation
+      }
+      return true;
+
+    case Op_VectorRearrange:
+      assert(bt != T_SHORT || VM_Version::supports_avx512bw(), "");
+      if (bt == T_BYTE && VM_Version::supports_avx512_vbmi()) {
+        return false; // Implementation limitation
+      }
+      return true;
```
XiaohongGong, line 1927: I guess the masked `VectorRearrange` is not generated currently, is it? So if AVX-512 supports the masked operations, will you add modify the API and intrinsics side for it in future?

jatin-bhateja: Yes, rearrange Template needs to accommodate extra mask argument, I added the backend changes in anticipation.

src/hotspot/share/opto/vectornode.cpp
jatin-bhateja: VectorNode::make_mask_node is only going to be called for masked operations. I don't think AndVMask is being created for non-predicated targets.  So we can take either approach.

src/hotspot/share/opto/vectorIntrinsics.cpp
jatin-bhateja: andNot and eq, but they are implemented in terms of and/or/xor/not. So its more of a optimization since AVX512 have direct instructions for andNot operations on mask.

src/hotspot/share/opto/vectornode.cpp
XiaohongGong: Ok, can the opcodes like `AndVMask` be blocked for non-predicated targets in `Matcher::match_rule_supported` ?

src/hotspot/share/opto/vectorIntrinsics.cpp
XiaohongGong: So I think this change is not needed, since the vector mask type will be created in the `VectorNode::make` method, which is currently used for `and/or/xor`.

src/hotspot/cpu/x86/x86.ad
XiaohongGong: OK, thank you for that. Since SVE doesn't support this masked operation, we will block it in SVE's backend.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8598,569 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+// ---------------------------------- Vector Masked Operations ------------------------------------
+
+instruct vadd_reg_masked(vec dst, vec src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (AddVB (Binary dst src2) mask));
+  match(Set dst (AddVS (Binary dst src2) mask));
+  match(Set dst (AddVI (Binary dst src2) mask));
+  match(Set dst (AddVL (Binary dst src2) mask));
+  match(Set dst (AddVF (Binary dst src2) mask));
+  match(Set dst (AddVD (Binary dst src2) mask));
```
sviswa7, line 8610: Could we add the match rules for sub, and, or, xor, mul, div, min, max, rearrange here?
The ins_encode part it the same for all of these.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8598,569 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+%}
+
+instruct vadd_mem_masked(vec dst, memory src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (AddVB (Binary dst src2) mask));
+  match(Set dst (AddVS (Binary dst src2) mask));
+  match(Set dst (AddVI (Binary dst src2) mask));
+  match(Set dst (AddVL (Binary dst src2) mask));
+  match(Set dst (AddVF (Binary dst src2) mask));
+  match(Set dst (AddVD (Binary dst src2) mask));
```
sviswa7, line 8629: Could we add the match rules for sub, and, or, xor, mul, div, min, max, rearrange here?
The ins_encode part it the same for all of these.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8598,569 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+                   $dst$$XMMRegister, $src2$$Address, true, vlen_enc);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct vlshift_reg_masked(vec dst, vec src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (LShiftVS (Binary dst src2) mask));
+  match(Set dst (LShiftVI (Binary dst src2) mask));
+  match(Set dst (LShiftVL (Binary dst src2) mask));
```
sviswa7, line 8833: Could we add the match rules for RShift, URShift here?
The ins_encode part it the same for all of these.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8598,569 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+                   $dst$$XMMRegister, $src2$$XMMRegister, true, vlen_enc, is_varshift);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct vlshift_mem_masked(vec dst, memory src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (LShiftVS (Binary dst src2) mask));
+  match(Set dst (LShiftVI (Binary dst src2) mask));
+  match(Set dst (LShiftVL (Binary dst src2) mask));
```
sviswa7, line 8850: Could we add the match rules for RShift, URShift here?
The ins_encode part it the same for all of these.

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -8897,7 +10385,7 @@ void Assembler::evex_prefix(bool vex_r, bool vex_b, bool vex_x, bool evex_r, boo
   // EVEX.b is not currently used for broadcast of single element or data rounding modes
   _attributes->set_evex_encoding(evex_encoding);
 
-  // P0: byte 2, initialized to RXBR`00mm
+  // P0: byte 2, initialized to RXBR00mm
```
sviswa7, line 10388: This should be  R'RXB00mm.

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7975,6 +8222,1247 @@ void Assembler::vpbroadcastw(XMMRegister dst, Address src, int vector_len) {
...
+  attributes.set_is_evex_instruction();
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
+  int encode = vex_prefix_and_encode(dst->encoding(), nds->encoding(), src->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int16(0x36, (0xC0 | encode));
+}
+
+void Assembler::evppermd(XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: evp is chosen as prefix in all the masked assemble routines. (evex + vector + packed)

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7975,6 +8222,1247 @@ void Assembler::vpbroadcastw(XMMRegister dst, Address src, int vector_len) {
...
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
+  vex_prefix(src, 0, dst->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int8(0x1F);
+  emit_operand(dst, src);
+}
+
+void Assembler::evfmaps(XMMRegister dst, KRegister mask, XMMRegister nds, XMMRegister src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: Existing implementation is not functionally incorrect, predication will still choose the old (dst) or new value, connections are set appropriately in IR.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Same as above

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Same as above

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Same as above

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: We will not be saving anything by clubbing the rules,  on pattern per operation is easy to review and maintain.
Your suggestion will make one pattern very bulky.

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7431,6 +7566,118 @@ void Assembler::evpxord(XMMRegister dst, KRegister mask, XMMRegister nds, XMMReg
   emit_int16((unsigned char)0xEF, (0xC0 | encode));
 }
 
+void Assembler::evpxord(XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: These are leaf level routines and existing assembler were lacking the routines with masked variant. Further re-factoring may complexify things?

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7431,6 +7566,118 @@ void Assembler::evpxord(XMMRegister dst, KRegister mask, XMMRegister nds, XMMReg
...
+
+void Assembler::evpxorq(XMMRegister dst, KRegister mask, XMMRegister nds, XMMRegister src, bool merge, int vector_len) {
+  assert(VM_Version::supports_evex(), "");
+  // Encoding: EVEX.NDS.XXX.66.0F.W1 EF /r
+  InstructionAttr attributes(vector_len, /* vex_w */ true, /* legacy_mode */ false, /* no_mask_reg */ false, /* uses_vl */ true);
+  attributes.set_is_evex_instruction();
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
```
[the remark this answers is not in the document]
jatin-bhateja: Merging/Zeroing semantics is associated with destination vector register.

src/hotspot/share/opto/vectornode.cpp
jatin-bhateja: Matcher::match_rule_supported only check the existence of pattern based on the opcode and default return value is true. 
AddVI v1 v2
AddVI v1 v2 mask

Where as match_rule_supported_vector_masked initially calls match_rule_supported, it then return false in default case and true only for operations for which masked patterns exist.

jatin-bhateja:
Hi @sviswa7 , Thanks for take a look at this. your comments have been addressed.

src/hotspot/cpu/x86/assembler_x86.cpp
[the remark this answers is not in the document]
sviswa7: Sounds good.

src/hotspot/cpu/x86/assembler_x86.cpp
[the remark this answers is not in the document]
sviswa7: ev is the prefix and evp.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+      fatal("Unexpected type argument %s", type2name(type));
+      break;
+  }
+}
+
+void MacroAssembler::kor(BasicType type, KRegister dst, KRegister src1, KRegister src2) {
+  switch(type) {
+    case T_BOOLEAN:
+    case T_BYTE:
+       korbl(dst, src1, src2);
```
sviswa7, line 8236: Is the type here the vector element type? Is so, for byte vector we need korql. For short kordl. For int korwl, For long korbl.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
   }
 }
 
+void MacroAssembler::kand(BasicType type, KRegister dst, KRegister src1, KRegister src2) {
+  switch(type) {
+    case T_BOOLEAN:
+    case T_BYTE:
+       kandbl(dst, src1, src2);
```
sviswa7, line 8212: Is the type here the vector element type? Is so, for byte vector we need kandql. For short kanddl. For int kandwl, For long kandbl.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+      fatal("Unexpected type argument %s", type2name(type));
+      break;
+  }
+}
+
+void MacroAssembler::kxor(BasicType type, KRegister dst, KRegister src1, KRegister src2) {
+  switch(type) {
+    case T_BOOLEAN:
+    case T_BYTE:
+       kxorbl(dst, src1, src2);
```
sviswa7, line 8260: Is the type here the vector element type? Is so, for byte vector we need kxorql. For short kxordl. For int kxorwl, For long kxorbl.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+      evpxorq(dst, mask, nds, src, merge, vector_len); break;
+    default:
+      fatal("Unexpected type argument %s", type2name(type)); break;
+  }
+}
+
+void MacroAssembler::evor(BasicType type, XMMRegister dst, KRegister mask, XMMRegister nds, XMMRegister src, bool merge, int vector_len) {
+  switch(type) {
+    case T_INT:
+      Assembler::evpord(dst, mask, nds, src, merge, vector_len); break;
```
sviswa7, line 8384: What about subword types here for or, xor, and?

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+    case T_INT:
+      evpandd(dst, mask, nds, src, merge, vector_len); break;
+    case T_LONG:
+      evpandq(dst, mask, nds, src, merge, vector_len); break;
+    default:
+      fatal("Unexpected type argument %s", type2name(type)); break;
+  }
+}
+
+void MacroAssembler::evpperm(BasicType type, XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
```
sviswa7, line 8425: This should be moved next to the register variant of the evpperm.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
jatin-bhateja: Skipped for now, since blend vector operation may be beneficial in that case instead of doing any special handling for sub-words.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
jatin-bhateja: No its not vector type. From the instruction side always T_LONG basic type is passed. TypeVectMask is vector of booleans.  We can use vector length to emit shorter instructions ,  but efficiency wise its wont be different to doing an operation over entire opmask register.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
jatin-bhateja: same as above

src/hotspot/cpu/x86/macroAssembler_x86.cpp
jatin-bhateja: Will do this change.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -3827,6 +3827,181 @@ void C2_MacroAssembler::arrays_equals(bool is_array_equ, Register ary1, Register
...
+    case Op_MinV:
+      evpmins(eType, dst, mask, src1, src2, merge, vlen_enc); break;
+    case Op_XorV:
+      evxor(eType, dst, mask, src1, src2, merge, vlen_enc); break;
+    case Op_OrV:
+      evor(eType, dst, mask, src1, src2, merge, vlen_enc); break;
+    case Op_AndV:
+      evand(eType, dst, mask, src1, src2, merge, vlen_enc); break;
+    case Op_AbsVD:
+    case Op_MulVB:
```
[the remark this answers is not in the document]
jatin-bhateja: Will remove them.

src/hotspot/share/opto/vector.cpp
```diff
@@ -338,7 +338,10 @@ Node* PhaseVector::expand_vbox_alloc_node(VectorBoxAllocateNode* vbox_alloc,
   int num_elem = vect_type->length();
 
   bool is_mask = is_vector_mask(box_klass);
-  if (is_mask && bt != T_BOOLEAN) {
+  // If boxed mask value is present in a predicate register, it must be
+  // spilled to a vector though a VectorStoreMaskOperation before actual StoreVector
+  // operation to vector payload field.
+  if (is_mask && (value->bottom_type()->isa_vectmask() || bt != T_BOOLEAN)) {
```
sviswa7, line 344: What happens on platforms that dont support predicate registers. Do we generate extra VectorStoreMask instruction now?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3335,7 +3450,18 @@ instruct sqrtD_reg(regD dst) %{
...
+instruct reinterpret_mask(kReg dst) %{
+  predicate(UseAVX > 2 && vector_length(n) == vector_length(n->in(1))); // dst == src
+  match(Set dst (VectorReinterpret dst));
+  ins_cost(125);
+  format %{ "vector_reinterpret $dst\t!" %}
+  ins_encode %{
+    // empty
+  %}
+  ins_pipe( pipe_slow );
+%}
```
sviswa7, line 3464: This is overriding the instructs reinterpret_expand* and reitnerpret_shrink. So doesnt look correct.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+    movzbl(dst, dst);
+  }
+}
+
+void MacroAssembler::alltrue(Register dst, uint masklen, KRegister src, KRegister kscratch) {
+  if (masklen < 8) {
+    kxnorbl(kscratch, kscratch, kscratch);
+    kshiftlbl(kscratch, kscratch, masklen);
+    korbl(kscratch, kscratch, src);
+    kortestbl(kscratch, kscratch);
```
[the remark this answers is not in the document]
jatin-bhateja: DONE

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -8205,6 +8205,312 @@ void MacroAssembler::evmovdqu(BasicType type, KRegister kmask, Address dst, XMMR
...
+      fatal("Unexpected type argument %s", type2name(type)); break;
+  }
+}
+
+void MacroAssembler::anytrue(Register dst, uint masklen, KRegister src, KRegister kscratch) {
+  if (masklen < 8) {
+    kxnorbl(kscratch, kscratch, kscratch);
+    kshiftrbl(kscratch, kscratch, 8-masklen);
+    kandbl(kscratch, kscratch, src);
+    ktestbl(kscratch, kscratch);
```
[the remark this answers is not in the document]
jatin-bhateja: DONE

src/hotspot/share/opto/vector.cpp
jatin-bhateja: This is as of now agnostic to existence of predication support. We dump the mask object into byte vector which is then used during object re-construction, this simplifies the re-construction logic but yes its an overhead for non-predicated targets. One area of improvement here even for predicated targets is to tie VectorStoreMask to same controlling edge as that of uncommon_trap call so that it does not get scheduled into higher ancestor block. I have taken a note of these points and will try to address in separate subsequent patch.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Both have different predication condition.

src/hotspot/share/opto/compile.cpp
```diff
@@ -2356,14 +2356,15 @@ bool Compile::has_vbox_nodes() {
...
+         n->req() == 2 &&
          VectorNode::is_vector_bitwise_not_pattern(n);
 }
 
 static bool is_vector_binary_bitwise_op(Node* n) {
   switch (n->Opcode()) {
     case Op_AndV:
     case Op_OrV:
-      return true;
+      return n->req() == 2;
```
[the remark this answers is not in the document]
jatin-bhateja: This is an interim solution for preventing macro logic optimization over masked logical operations. I have a separate patch for masked macro logic inferencing.

jatin-bhateja:
Hi @sviswa7 , your comments have been addressed. Thanks

src/hotspot/cpu/x86/x86.ad
```diff
@@ -6993,26 +7175,56 @@ instruct evcmp(vec dst, vec src1, vec src2, immI8 cond, rRegP scratch, kReg ktmp
...
+        break;
+      }
+      default: assert(false, "%s", type2name(src1_elem_bt));
+    }
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+
+instruct evcmp(kReg dst, vec src1, vec src2, immI8 cond, rRegP scratch) %{
```
sviswa7, line 7196: scratch register is not used.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -6920,8 +7082,28 @@ instruct evcmpFD(vec dst, vec src1, vec src2, immI8 cond, rRegP scratch, kReg kt
   ins_pipe( pipe_slow );
 %}
 
+instruct evcmpFD(kReg dst, vec src1, vec src2, immI8 cond, rRegP scratch) %{
```
sviswa7, line 7085: scratch register is not used.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3632,7 +3758,7 @@ instruct gather(legVec dst, memory mem, legVec idx, rRegP tmp, legVec mask) %{
 %}
 
 instruct evgather(vec dst, memory mem, vec idx, rRegP tmp, kReg ktmp) %{
-  predicate(vector_length_in_bytes(n) == 64);
+  predicate(VM_Version::supports_avx512vl() || vector_length_in_bytes(n) == 64);
   match(Set dst (LoadVectorGather mem idx));
   effect(TEMP dst, TEMP tmp, TEMP ktmp);
   format %{ "load_vector_gather $dst, $mem, $idx\t! using $tmp and k2 as TEMP" %}
```
sviswa7, line 3764: There is no k2 here, should be ktmp.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3651,6 +3777,23 @@ instruct evgather(vec dst, memory mem, vec idx, rRegP tmp, kReg ktmp) %{
   ins_pipe( pipe_slow );
 %}
 
+instruct evgather_masked(vec dst, memory mem, vec idx, kReg mask, kReg ktmp, rRegP tmp) %{
+  predicate(VM_Version::supports_avx512vl());
+  match(Set dst (LoadVectorGatherMasked mem (Binary idx mask)));
+  effect(TEMP_DEF dst, TEMP tmp, TEMP ktmp);
+  format %{ "load_vector_gather_masked $dst, $mem, $idx, $mask\t! using $tmp and k2 as TEMP" %}
```
sviswa7, line 3784: There is no k2 here, should be ktmp.

src/hotspot/cpu/x86/x86.ad
sviswa7: ok

src/hotspot/cpu/x86/x86.ad
```diff
@@ -7402,57 +7662,101 @@ instruct vptest_anytrue(rRegI dst, legVec src1, legVec src2, rFlagsReg cr) %{
...
+    __ ktestbl($ktmp$$KRegister, $ktmp$$KRegister);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct cmpvptest_anytrue_ge8_evex(rFlagsReg cr, kReg src1, kReg src2, immI_0 zero) %{
+  predicate(VM_Version::supports_avx512bwdq() &&
+            static_cast<const VectorTestNode*>(n->in(1))->get_predicate() == BoolTest::ne &&
+            vector_length(n->in(1)->in(1)) >= 8);
+  match(Set cr (CmpI (VectorTest src1 src2) zero));
```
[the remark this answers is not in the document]
jatin-bhateja: Yes, this is missing in mainline code too. Taken a note of this, will be added in subsequent patch.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -7621,10 +7898,39 @@ instruct storeMask8B_avx(vec dst, vec src, immI_8 size, legVec vtmp) %{
...
+      src_vlen_enc = Assembler::AVX_512bit;
+    }
+    __ evpmovdb($dst$$XMMRegister, $src$$XMMRegister, src_vlen_enc);
+    __ vpabsb($dst$$XMMRegister, $dst$$XMMRegister, dst_vlen_enc);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct vstoreMask8B_evex(vec dst, vec src, immI_8 size) %{
+  predicate(UseAVX > 2 && NULL == n->in(1)->bottom_type()->isa_vectmask());
```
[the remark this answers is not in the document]
jatin-bhateja: DONE, in general I follow CONST == VAR (lvalue) convention to avoid accidental assignment in comparison condition (VAR = CONST)

src/hotspot/cpu/x86/x86.ad
```diff
@@ -7637,7 +7943,21 @@ instruct vstoreMask8B_evex(vec dst, vec src, immI_8 size) %{
...
+instruct vstoreMask_evex(vec dst, kReg mask, immI size, rRegP scratch) %{
+  predicate(UseAVX > 2 && n->in(1)->bottom_type()->isa_vectmask());
+  match(Set dst (VectorStoreMask mask size));
+  effect(TEMP scratch);
+  format %{ "vector_store_mask $dst,$mask\t!" %}
+  ins_encode %{
+    int dst_vlen_enc = vector_length_encoding(this);
+    __ vpxor($dst$$XMMRegister, $dst$$XMMRegister, $dst$$XMMRegister, dst_vlen_enc);
+    __ evmovdqub($dst$$XMMRegister, $mask$$KRegister, ExternalAddress(vector_masked_cmp_bits()),
+                  true, dst_vlen_enc, $scratch$$Register);
```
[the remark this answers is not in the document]
jatin-bhateja: Yes, proposed sequence  though have more number of instructions but will also prevent any penalty due to cache miss.
existing:
  VCMP  K1  MASK_VEC, MEM_BIT_PATTERN
Proposed:
  VXOR  V1 V1
  VSUB  V1 , MASK_VEC
  VMOVM2B   K1 , V1
 I did quick perf runs and did see increased instruction could

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+instruct vrearrangev_mem_masked(vec dst, memory src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (VectorRearrange (Binary dst src2) mask));
+  format %{ "vprearrange_masked $dst, $dst, $src2\t! rearrange masked operation" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(this);
+    BasicType bt = vector_element_basic_type(this);
+    int opc = this->ideal_Opcode();
+    __ evmasked_op(opc, bt, $mask$$KRegister, $dst$$XMMRegister,
+                   $dst$$XMMRegister, $src2$$Address, false, vlen_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: same as above

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+instruct vrearrangev_reg_masked(vec dst, vec src2, kReg mask) %{
+  predicate(UseAVX > 2);
+  match(Set dst (VectorRearrange (Binary dst src2) mask));
+  format %{ "vprearrange_masked $dst, $dst, $src2\t! rearrange masked operation" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(this);
+    BasicType bt = vector_element_basic_type(this);
+    int opc = this->ideal_Opcode();
+    __ evmasked_op(opc, bt, $mask$$KRegister, $dst$$XMMRegister,
+                   $dst$$XMMRegister, $src2$$XMMRegister, false, vlen_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: All these instructions are in two address format, thus src1 is copied to dst first.  Second flavor of rearrange does not accept mask argument. This pattern specifically handled the case with mask.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+  ins_pipe( pipe_slow );
+%}
+
+instruct vabs_masked(vec dst, kReg mask) %{
+  match(Set dst (AbsVB dst mask));
+  match(Set dst (AbsVS dst mask));
+  match(Set dst (AbsVI dst mask));
+  match(Set dst (AbsVL dst mask));
+  format %{ "vabs_masked $dst, $mask \t! vabs masked operation" %}
+  ins_cost(100);
```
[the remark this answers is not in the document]
jatin-bhateja: To give this preference over Set dst (AbsVI src) pattern

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+  predicate(UseAVX > 2);
+  match(Set dst (FmaVF (Binary dst src2) (Binary src3 mask)));
+  match(Set dst (FmaVD (Binary dst src2) (Binary src3 mask)));
+  format %{ "vfma_masked $dst, $src2, $src3, $mask \t! vfma masked operation" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(this);
+    BasicType bt = vector_element_basic_type(this);
+    int opc = this->ideal_Opcode();
+    __ evmasked_op(opc, bt, $mask$$KRegister, $dst$$XMMRegister,
+                   $src2$$XMMRegister, $src3$$Address, true, vlen_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: same as above

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7975,6 +8222,1218 @@ void Assembler::vpbroadcastw(XMMRegister dst, Address src, int vector_len) {
...
+  int encode = vex_prefix_and_encode(dst->encoding(), nds->encoding(), src->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int16((unsigned char)0xA8, (0xC0 | encode));
+}
+
+void Assembler::evpfma213pd(XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: same as above

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7975,6 +8222,1218 @@ void Assembler::vpbroadcastw(XMMRegister dst, Address src, int vector_len) {
...
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
+  vex_prefix(src, nds->encoding(), dst->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int8((unsigned char)0xA8);
+  emit_operand(dst, src);
+}
+
+void Assembler::evpfma213pd(XMMRegister dst, KRegister mask, XMMRegister nds, XMMRegister src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: same as above

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7975,6 +8222,1218 @@ void Assembler::vpbroadcastw(XMMRegister dst, Address src, int vector_len) {
...
+  attributes.set_is_evex_instruction();
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
+  int encode = vex_prefix_and_encode(dst->encoding(), nds->encoding(), src->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int16((unsigned char)0xA8, (0xC0 | encode));
+}
+
+void Assembler::evpfma213ps(XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
```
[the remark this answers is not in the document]
jatin-bhateja: This is conforming with generated IR shape, dst is first source i.e. src1 followed by src2  and  src3
FMA213 = src1*src2 + src3.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP scratch) %{
+  predicate(UseAVX > 2);
```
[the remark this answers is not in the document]
jatin-bhateja: That is already checked in match_rule_supported_vector_masked. Infact UseAVX > 2 is also redundant in all newly added patterns.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+%}
+
+instruct mask_all_evexI(kReg dst, rRegI src, vec xtmp) %{
+  match(Set dst (MaskAll src));
+  effect(TEMP xtmp);
+  format %{ "mask_all_evexI $dst, $src \t! mask all operation" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(vector_length(this));
+    __ evpbroadcastb($xtmp$$XMMRegister, $src$$Register, vlen_enc);
+    __ evpmovb2m($dst$$KRegister, $xtmp$$XMMRegister, vlen_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: same as above

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8269,6 +8601,570 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+
+#ifdef _LP64
+instruct mask_all_evexL(kReg dst, rRegL src, vec xtmp) %{
+  match(Set dst (MaskAll src));
+  effect(TEMP xtmp);
+  format %{ "mask_all_evexL $dst, $src \t! mask all operation" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(vector_length(this));
+    __ evpbroadcastb($xtmp$$XMMRegister, $src$$Register, vlen_enc);
+    __ evpmovb2m($dst$$KRegister, $xtmp$$XMMRegister, vlen_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: This is a mask all operation, mask (src) is initially only 0/-1,  this needs to be broadcasted into vector and then movb2m picks the MSB bits from each byte lane and moves it to opmask register (dst).

src/hotspot/share/opto/vectornode.cpp
XiaohongGong: But I think `AndVMask/OrVMask/XorVMask` are different from the masked vector opcodes like `AddVI`. I was intending to make nodes like `AndVMask/OrVMask/XorVMask` also work for the non-mask supported platforms. Other platforms can also support them by adding the backend implementations if they want. We can just treat them as a normal vector operations for vector mask.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -9065,34 +9036,47 @@ instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP
...
+  effect(TEMP_DEF dst, TEMP tmp);
   format %{ "mask_all_evexI $dst, $src \t! mask all operation" %}
   ins_encode %{
-    int vlen_enc = vector_length_encoding(vector_length(this));
-    __ evpbroadcastb($xtmp$$XMMRegister, $src$$Register, vlen_enc);
-    __ evpmovb2m($dst$$KRegister, $xtmp$$XMMRegister, vlen_enc);
+    int vec_len = vector_length(this);
+    __ movslq($tmp$$Register, $src$$Register);
+    __ kmovql($dst$$KRegister, $tmp$$Register);
+    __ kshiftrql($dst$$KRegister, $dst$$KRegister, 64 - vec_len);
```
sviswa7, line 9061: Could we not do kmovdl followed by kshiftrdl here? Thereby removing the need for movslq.

src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
```diff
@@ -4001,6 +4001,7 @@ class StubGenerator: public StubCodeGenerator {
     StubRoutines::x86::_vector_byte_perm_mask = generate_vector_byte_perm_mask("vector_byte_perm_mask");
     StubRoutines::x86::_vector_long_sign_mask = generate_vector_mask_long_double("vector_long_sign_mask", 0x80000000, 0x00000000);
     StubRoutines::x86::_vector_all_bits_set = generate_vector_mask("vector_all_bits_set", 0xFFFFFFFF);
+    StubRoutines::x86::_vector_masked_cmp_bits = generate_vector_mask("vector_masked_cmp_bits", 0x01010101);
```
sviswa7, line 4004: _vector_masked_cmp_bits can be removed now.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -7474,24 +7778,25 @@ instruct loadMask(legVec dst, legVec src) %{
...
   format %{ "vector_loadmask_byte $dst,$src\n\t" %}
   ins_encode %{
-    int vlen_in_bytes = vector_length_in_bytes(this);
-    BasicType elem_bt = vector_element_basic_type(this);
-
-    __ load_vector_mask($dst$$XMMRegister, $src$$XMMRegister, vlen_in_bytes, elem_bt, false);
+    int vlen_in_bytes = vector_length_in_bytes(in(1));
+    int vlen_enc = vector_length_encoding(vlen_in_bytes);
+    __ evpcmp(T_BYTE, $dst$$KRegister, k0, $src$$XMMRegister, ExternalAddress(vector_masked_cmp_bits()),
+              Assembler::eq, vlen_enc, $scratch$$Register);
```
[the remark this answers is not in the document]
sviswa7: This is still pending.

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
sviswa7: Could you clarify further please. e.g. vpermd is a 3 address instruction:
 VPERMD zmm1 {k1}{z}, zmm2, zmm3/m512/m32bcst :Permute doublewords in zmm3/m512/m32bcst using indices in zmm2 and store the result in zmm1 using writemask k1.

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
sviswa7: The pattern is different, one has mask and one doesn't.
match(Set dst (AbsVI dst mask));
match(Set dst (AbsVI src));

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
sviswa7: In non masked version, fma231 is being used.  Please double check:
FMA operation is dst = dst + src2 * src3, where dst is the accumulator.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3674,6 +3825,23 @@ instruct scatter(memory mem, vec src, vec idx, rRegP tmp, kReg ktmp) %{
...
+  predicate(VM_Version::supports_avx512vl());
+  match(Set mem (StoreVectorScatterMasked mem (Binary src (Binary idx mask))));
+  effect(TEMP tmp, TEMP ktmp);
+  format %{ "store_vector_scatter_masked $mem, $idx, $src, $mask\t!" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(this, $src);
+    BasicType elem_bt = vector_element_basic_type(this, $src);
+    assert(vector_length_in_bytes(this, $src) >= 16, "sanity");
+    assert(!is_subword_type(elem_bt), "sanity"); // T_INT, T_LONG, T_FLOAT, T_DOUBLE
+    __ kmovwl($ktmp$$KRegister, $mask$$KRegister);
```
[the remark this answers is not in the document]
jatin-bhateja: Gather/Scatter instruction partially updates predicate register, hence moving mask to temporary.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -3651,6 +3785,23 @@ instruct evgather(vec dst, memory mem, vec idx, rRegP tmp, kReg ktmp) %{
...
+  predicate(VM_Version::supports_avx512vl());
+  match(Set dst (LoadVectorGatherMasked mem (Binary idx mask)));
+  effect(TEMP_DEF dst, TEMP tmp, TEMP ktmp);
+  format %{ "load_vector_gather_masked $dst, $mem, $idx, $mask\t! using $tmp and ktmp as TEMP" %}
+  ins_encode %{
+    assert(UseAVX > 2, "sanity");
+    int vlen_enc = vector_length_encoding(this);
+    BasicType elem_bt = vector_element_basic_type(this);
+    assert(!is_subword_type(elem_bt), "sanity"); // T_INT, T_LONG, T_FLOAT, T_DOUBLE
+    __ kmovwl($ktmp$$KRegister, $mask$$KRegister);
```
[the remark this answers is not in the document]
jatin-bhateja: Gather/Scatter instruction partially updates predicate register, hence moving mask to temporary.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -1845,7 +1853,122 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType
...
+      }
+      return true;
+
+    // Binary Logical operations
+    case Op_AndVMask:
+    case Op_OrVMask:
+    case Op_XorVMask:
+      assert(bt != T_INT  || VM_Version::supports_avx512bw(), "");
+      assert(bt != T_LONG || VM_Version::supports_avx512bw(), "");
+      if (bt == T_BYTE && VM_Version::supports_avx512dq()) {
```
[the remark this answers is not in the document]
jatin-bhateja: DONE

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: maskAll accept a boolean argument (false (0) , true(-1)).
value operand has a rRegI register class which represent 32 bit register. This value need to be sign extended to 64 bit value before computing the final mask value using shift right operation.

jatin-bhateja:
Hi @sviswa7 , your comments have been addressed
Along with the latest merge, backend support for masked memory operations over byteArray and byteBuffers has also been added. New patterns for VectorReinterpret operation for mask types and minor IR change related to it.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -7192,9 +7252,9 @@ instruct vcmpu32(legVec dst, legVec src1, legVec src2, immI8 cond, legVec vtmp1,
 
 instruct vcmpu64(vec dst, vec src1, vec src2, immI8 cond, rRegP scratch, kReg ktmp) %{
   predicate(UseAVX > 2 &&
-            !VM_Version::supports_avx512vl() &&
-            vector_length_in_bytes(n->in(1)->in(1)) == 64 && // src1
-            is_integral_type(vector_element_basic_type(n->in(1)->in(1)))); // src1
+            (VM_Version::supports_avx512vl() ||
+             Matcher::vector_length_in_bytes(n->in(1)->in(1)) == 64) && // src1
```
sviswa7: This doesn't look correct. This instruct sets vec dst, whereas with avx512vl kReg dst is expected.
Should have been:
            !VM_Version::supports_avx512vl() &&
            vector_length_in_bytes(n->in(1)->in(1)) == 64 &&

sviswa7:
Other than the predicate above for vcmpu64, rest of the patch looks good to me. 
Please fix the predicate. No need for re-review.

sviswa7 (approved):

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Yes this new pattern was added specifically for 64 byte unsigned/signed comparisons (will apply to integral types only) for targets which do not support VL extensions.

PR #104, opened 2021-07-27
Optimize mem barriers for ByteBuffer cases

rsmogura:
# Description
This change tries to remove mem bars for byte buffer cases.

Previously mem bars were inserted almost unconditionally if attemp to native memory acees where detected. This patch tries to follow up inline_unsafe_access and insert bar only if can't determine if it's heap or off-heap (type missmatch cases are not ported).

# Testing
Memory tests should include rollbacking JDK changes, and leaving only hotspot, as intrinsics should be well guarded

# Notes
Polluted cases to be addressed later

# Benchmarks
```
Benchmark                                (size)  Mode  Cnt    Score   Error  Units
ByteBufferVectorAccess.arrays              1024  avgt   10   12.585 ? 0.409  ns/op
ByteBufferVectorAccess.directBuffers       1024  avgt   10   19.962 ? 0.080  ns/op
ByteBufferVectorAccess.heapBuffers         1024  avgt   10   15.878 ? 0.187  ns/op
ByteBufferVectorAccess.pollutedBuffers2    1024  avgt   10  123.702 ? 0.723  ns/op
ByteBufferVectorAccess.pollutedBuffers3    1024  avgt   10  223.928 ? 1.906  ns/op
```
Before
```
Benchmark                                (size)  Mode  Cnt    Score   Error  Units
ByteBufferVectorAccess.arrays              1024  avgt   10   14.730 ? 0.061  ns/op
ByteBufferVectorAccess.directBuffers       1024  avgt   10   77.707 ? 4.867  ns/op
ByteBufferVectorAccess.heapBuffers         1024  avgt   10   76.530 ? 1.076  ns/op
ByteBufferVectorAccess.pollutedBuffers2    1024  avgt   10  143.331 ? 1.096  ns/op
ByteBufferVectorAccess.pollutedBuffers3    1024  avgt   10  286.645 ? 3.444  ns/op
```

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/104/head:pull/104` \
`$ git checkout pull/104`

`$ git checkout pull/104` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/104/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 104`

`$ git pr show -t 104`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/104.diff">https://git.openjdk.java.net/panama-vector/pull/104.diff</a>

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,7 @@
 /build/
 /dist/
 /.idea/
-/**/.idea
-/**/*.iml
 /.vscode/
-/**/.vscode
 /nbproject/
 nbproject/private/
 /webrev
@@ -21,7 +18,3 @@ NashornProfile.txt
 /src/utils/LogCompilation/target/
 /.project/
 /.settings/
-
-# Downloaded binaries
-/src/utils/hsdis/binutils*
-/src/utils/hsdis/build/
--- a/src/.idea/.gitignore
+++ b/src/.idea/.gitignore
@@ -1,3 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml


PaulSandoz:
This is starting to look better, the logic for heap/native looks correct comparing with unsafe. Perhaps for clarity we should use `DecoratorSet`?

It got me wondering more why we did not have to do similar updates in Java code for VH accesses to byte buffers. Your `ByteBufferBarriersTests` benchmark indicates VHs work for non-polluted heap and direct cases.

I played around with your patch and i managed to simplify by reducing the code changes in scoped memory access classes to a cast of the base to `byte[]`, e.g. src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template

```
    @Scoped
    @ForceInline
    private static
    <V extends VectorSupport.Vector<E>, E, S extends VectorSupport.VectorSpecies<E>>
    V loadFromByteBufferScoped(ScopedMemoryAccess.Scope scope,
                          Class<? extends V> vmClass, Class<E> e, int length,
                          ByteBuffer bb, int offset,
                          S s,
                          VectorSupport.LoadOperation<ByteBuffer, V, E, S> defaultImpl) {
        try {
            if (scope != null) {
                scope.checkValidState();
            }

            return VectorSupport.load(vmClass, e, length,
                (byte[]) BufferAccess.bufferBase(bb), BufferAccess.bufferAddress(bb, offset),
                bb, offset, s,
                defaultImpl);
        } finally {
            Reference.reachabilityFence(scope);
        }
    }
```
 
I don't know why this works.

I am also observing an odd performance issue with heap access when tiered compilation is enabled, it may be unrelated as i can reproduce using the tip of `jdk/jdk`.

rsmogura:
To use decorators I would  move C2AccessFence to hpp file (right now it's CPP private) - this class actually is reponsible for setting up barriers. I think this can be a good idea.

So, with polluted access there are actually two things which can happen - both resulting in same performance degradation.

In polluted case, If we will have loadFromByteBufferScoped without "if" , the intrinsic will not be able to detect if access is native or not and will start putting barriers.

With "if" enabled - there's a less optimized graph generated by hotspot. Mainly "if" causes memory merge which is marked as bot, reading of hb, limit fields takes memory from this merge and phi - in perfect world should not as merge does not affect this slice. As a consequence this fields are read with every loop pass - and are not pulled outside loop.

I think I already found a good (or at least) working idea how to overcome this purely in Java. 

Funny thing I've found a code in hotspot which theoretically can help, but it requires tunning [1].

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/cfgnode.cpp#L2212

rsmogura:
I did slighlty bigger refactor and pulled out logic responsible for setting decorators from library_call.

PaulSandoz:
What if we use unsafe to read the BB read only field? we do that for VHs:
```
        @ForceInline
        static int indexRO(ByteBuffer bb, int index) {
            if (UNSAFE.getBoolean(bb, BYTE_BUFFER_IS_READ_ONLY))
                throw new ReadOnlyBufferException();
            return index(bb, index);
        }
```
The virtual method call is complicating matters.

Logistically, having seen the scope of changes, perhaps it's best to first focus on a smaller changeset for the specific problem, then revisit with better use of the decorators? That way we could integrate this PR to `vectorIntrinsics` then bring into `jdk`  then afterwards decide on a refactor, with input from the hotspot reviewers?

rsmogura:
> What if we use unsafe to read the BB read only field? we do that for VHs:
> 
> ```
>         @ForceInline
>         static int indexRO(ByteBuffer bb, int index) {
>             if (UNSAFE.getBoolean(bb, BYTE_BUFFER_IS_READ_ONLY))
>                 throw new ReadOnlyBufferException();
>             return index(bb, index);
>         }
> ```
> 
> The virtual method call to `Buffer.isReadOnly` might be complicating matters.
> 
> Logistically, having seen the scope of changes, perhaps it's best to first focus on a smaller changeset for the specific problem, then revisit with better use of the decorators? That way we could integrate this PR to `vectorIntrinsics` then bring into `jdk` then afterwards decide on a refactor, with input from the hotspot reviewers?

Yes, this change looks huge... a bit or even two bits...

I think it's a very good idea to use unsafe with BB to check for read only, that's something I was missing. However, outermost "if" should be a check if a buffer is direct or not and it should take input from immutable memory this way we have more chances to unswitch loop on this what creates the biggest issue: splitting load/store to on-heap or off-heap (otherwise in polluted case, it can be problematic). Here, we should use instanceof, instead of isDirect, as isDirect is virtual call and in most polluted cases this can kick in (unless we can somehow optimize isDirect)

rsmogura:
Hi all, I think that for this PR, I rather don't want to push more changes. It already stepped out from the original concern of optimizing mem-bars, to optimizing polluted cases.

PaulSandoz:
> Hi all, I think that for this PR, I rather don't want to push more changes. It already stepped out from the original concern of optimizing mem-bars, to optimizing polluted cases.

Yeah, I am inclined to dial back to focus this PR on just the non-polluted cases. 
AFAICT I think we can get a set of minimal changes using casting and unsafe for accessing buffer state with no need to modify byte buffer code or branch on heao/direct. WDYT?

We can then expand later with further experiments for polluted cases, evaluating the costs and benefits.

rsmogura:
> > Hi all, I think that for this PR, I rather don't want to push more changes. It already stepped out from the original concern of optimizing mem-bars, to optimizing polluted cases.
> 
> Yeah, I am inclined to dial back to focus this PR on just the non-polluted cases.
> AFAICT I think we can get a set of minimal changes using casting and unsafe for accessing buffer state with no need to modify byte buffer code or branch on heao/direct. WDYT?
> 
> We can then expand later with further experiments for polluted cases, evaluating the costs and benefits.

Definitely, to all of the above (branching actually was added for polluted cases) - this went too big.

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
```diff
@@ -406,10 +406,19 @@ public class ScopedMemoryAccess {
...
+              return VectorSupport.load(vmClass, e, length,
+                      null, BufferAccess.bufferAddress(bb, offset),
+                      bb, offset, s,
+                      defaultImpl);
+            } else {
+              return VectorSupport.load(vmClass, e, length,
+                      base, BufferAccess.bufferAddress(bb, offset),
+                      bb, offset, s,
+                      defaultImpl);
+            }
```
PaulSandoz, line 421: Can we try collapsing that to:
```suggestion
              return VectorSupport.load(vmClass, e, length,
                      base, BufferAccess.bufferAddress(bb, offset),
                      bb, offset, s,
                      defaultImpl);
```
? That appeared to be ok when i ran the benchmark for the non-polluted cases.

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -760,12 +760,29 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
...
+  const bool can_access_non_heap = TypePtr::NULL_PTR->higher_equal(base_type);
+
+  // Not determined access, base can and can not be null.
+  const bool mixed_access = !off_heap_access && can_access_non_heap;
+  bool mismatched_array = false;
 
   const TypePtr *addr_type = gvn().type(addr)->isa_ptr();
   const TypeAryPtr* arr_type = addr_type->isa_aryptr();
 
+  if (arr_type == NULL && !off_heap_access) {
```
PaulSandoz, line 781: Not sure this case can occur, since we only pass array instances or `null` and we cannot pollute with different types of array. Might need some guidance from a HotSpot expert? Perhaps we can remove this code or place in an assert?

We do have a form of mismatched access between `char[]` and `ShortVector` and `boolean[]` and `ByteVector` but that's a different case.

test/hotspot/jtreg/compiler/vectorapi/VectorMemoryAlias.java
```diff
@@ -0,0 +1,68 @@
...
+        throw new AssertionError();
+      }
+    }
+  }
+
+  public static int test() {
+    byte arr[] = new byte[256];
+    final var bb = ByteBuffer.wrap(arr);
+    final var vs = VectorSpecies.ofLargestShape(byte.class);
+    final var ones = ByteVector.broadcast(vs, 1);
```
PaulSandoz, line 58: Make these static finals (or at least the former)?

test/micro/org/openjdk/bench/jdk/incubator/vector/ByteBufferVectorAccess.java
```diff
@@ -0,0 +1,152 @@
...
+    byte[] out = heapOut.array();
+
+    for (int i=0; i < SPECIES.loopBound(in.length); i += SPECIES.vectorByteSize()) {
+      final var v = ByteVector.fromArray(SPECIES, in, i);
+      v.intoArray(out, i);
+    }
+  }
+
+  @CompilerControl(CompilerControl.Mode.DONT_INLINE)
+  protected void copyMemory(ByteBuffer in, ByteBuffer out) {
```
PaulSandoz, line 146: We may want to create a wrapping method that is not inlined just for the polluted cases e.g. `copyIntoNotInlined` calling `copyMemory`?

Thereby we avoid the measurement of a method call in the non-polluted cases so as to compare more fairly with `arrayCopy`?

PaulSandoz (commented):
Appreciate all your efforts and explorations. We have been at this for a few rounds and it looks like we are getting close.

Two things:
1. Did you ever get the `VectorMemoryAlias` test to fail e.g. by deliberately disabling the barriers?
2. I think we should use unsafe to test for a read only buffer. We can follow up with another PR. (My hope is this might can also resolve the issue with tiered compilation and heap buffers i was observing.)

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
rsmogura: When I reverted changes I took base as `Object`, from `BufferAccess`, and did not work. Looks like cast to byte[] implies some kind of null checks, and we can use this.

rsmogura:
Hi @PaulSandoz, thank you for your suggestions and please check a new revision.

The test `VectorMemoryAlias` will fail if barriers will be removed in C++ code (and we have as well drop changes in `X-ScopedMemoryAccess` to avoid null checks).

---

I'm checking as well another approach, which I came out with today.

The purpose of barriers is to "refresh" memory in case of mixed access. However what if in case of mixed access, we would do:
* for a store we could assign result of StoreVector to two slices raw, and byte[] in a memory merge node,
* for a load, we could consume the whole memory as input, instead of a single slice.

Do you think this makes sense?

[1] https://github.com/rsmogura/panama-vector/commits/merege-mem-instead-of-barier

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -823,7 +834,7 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
 
   const TypeInstPtr* vbox_type = TypeInstPtr::make_exact(TypePtr::NotNull, vbox_klass);
 
-  if (can_access_non_heap) {
+  if (mixed_access) {
```
iwanowww, line 837: I suggest the following shapes for the checks:
```
const bool in_native = (TypePtr::NULL_PTR == base_type);            // base always null
const bool in_heap   = !TypePtr::NULL_PTR->higher_equal(base_type); // base never null

const bool is_mixed_access = !in_heap && !in_native;

const bool is_mismatched_access = in_heap && (addr_type->isa_aryptr() == NULL);

const bool needs_cpu_membar = is_mixed_access || is_mismatched_access;
```

iwanowww (changes requested):
`.gitignore` changes are unrelated.

You reverted mismatched access checks. Please, bring them back.

iwanowww:
> for a store we could assign result of StoreVector to two slices raw, and byte[] in a memory merge node,

I don't see how it could work with the alias analysis (as it is implemented now). 
Every memory slice is "flattened" into a unique slice which doesn't alias with anything except the one represented with `TypePtr::BOTTOM`. What you suggest implies that some slices start to alias with raw memory. It will break the existing logic unless you find a smart way to fix it.   

> for a load, we could consume the whole memory as input, instead of a single slice.

Still, you need to be very cautious about the alias index being assigned to the "wide" memory slice of mixed/mismatched access.  Also, the logic which inserts anti-dependencies in the graph has to be taught about the aliasing slices. 

Overall, it looks error-prone and it wouldn't necessarily lead to a simpler IR (and better generated code) compared to CPU membars.

src/hotspot/share/opto/vectorIntrinsics.cpp
rsmogura: Thank you, fixed, we were not sure if mismatched access should be there.

iwanowww (approved):
Looks good.

PaulSandoz (approved):
Very good. After integration i think we can provide a PR to the jdk repository.

rsmogura:
> Very good. After integration i think we can provide a PR to the jdk repository.

Thank you all. And sorry for causing maybe too much bother.

PR #106, opened 2021-08-03
Optimize mem barriers hotspot plus

rsmogura:
This is an experiment to assigne loads accessors bypassing other nodes. Sometimes loop invariant nodes "is hidden" behind a  chain of phi and merges, but this chain does not modify the loaded memory slice, and node could be finally anchored at start.

Such node is not seen as loop invariant, and prevents range check elimination, and other optimizations,

During build this code gets executed on few classes in JDK, so it's not vector specific.

The class is called LoadOptimize. PR is mix of changes from the original PR, but is enriched with load optimizer

I wonder if this have a sense?

Benchamrks gives same results as in original PR, but code is simplified
```
Benchmark                                (size)  Mode  Cnt   Score   Error  Units
ByteBufferVectorAccess.arrayCopy           1024  avgt   10  14.499 ? 0.267  ns/op
ByteBufferVectorAccess.directBuffers       1024  avgt   10  13.064 ? 0.115  ns/op
ByteBufferVectorAccess.heapBuffers         1024  avgt   10  16.524 ? 2.069  ns/op
ByteBufferVectorAccess.pollutedBuffers2    1024  avgt   10  34.459 ? 0.461  ns/op
ByteBufferVectorAccess.pollutedBuffers3    1024  avgt   10  66.688 ? 1.423  ns/op
ByteBufferVectorAccess.pollutedBuffers4    1024  avgt   10  61.904 ? 2.652  ns/op
ByteBufferVectorAccess.pollutedBuffers5    1024  avgt   10  61.967 ? 1.862  ns/op
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/106/head:pull/106` \
`$ git checkout pull/106`

`$ git checkout pull/106` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/106/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 106`

`$ git pr show -t 106`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/106.diff">https://git.openjdk.java.net/panama-vector/pull/106.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/nio/Buffer.java
+++ b/src/java.base/share/classes/java/nio/Buffer.java
@@ -797,12 +797,25 @@ public abstract class Buffer {
 
                 @Override
                 public Object getBufferBase(ByteBuffer bb) {
+                    assert 1==2;
                     return bb.base();
+                }
+                @Override
+                @ForceInline
+                public byte[] getBufferBaseFiled(ByteBuffer bb) {
+                    return bb.hb;
                 }
 
                 @Override
+                @ForceInline
                 public long getBufferAddress(ByteBuffer bb) {
                     return bb.address;
+                }
+
+                @Override
+                @ForceInline
+                public boolean isDirect(ByteBuffer bb) {
+                    return !(bb instanceof HeapByteBuffer);
                 }
 
                 @Override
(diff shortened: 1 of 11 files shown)


rsmogura:
I opened this accidently - I don't think it's correct. Sorry.

PR #107, opened 2021-08-04
8271539: Add masking support for load/store from/into byte array/buffer

XiaohongGong:
This patch adds the predicate support for the following masked vector load/store operations:
  - fromByteArray/fromByteBuffer
  - intoByteArray/intoByteBuffer

The masked APIs can respectively call the `VectorSupport.loadMasked/storeMasked` intrinsics with transforming the right mask class and value to them. The hotspot compiler will check whether it is using byte array/buffer when doing the memory accessing, and the input vector mask should be reinterpreted to a mask with the byte element type before used if it is. Besides, similar with the non-masked operations, the input vector to store should be reinterpreted to byte type before storing, and the vector loaded from the byte memory should be reinterpreted to the real vector type.

Also fix an issue when checking array indexes with mask before calling the intrinsics. The original checking method doesn't take effect for negative offset.

Note that this patch doesn't contain any backend changes.

---------
### Progress

### Issue
 * [JDK-8271539](https://bugs.openjdk.java.net/browse/JDK-8271539): Add masking support for load/store from/into byte array/buffer

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/107/head:pull/107` \
`$ git checkout pull/107`

`$ git checkout pull/107` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/107/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 107`

`$ git pr show -t 107`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/107.diff">https://git.openjdk.java.net/panama-vector/pull/107.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -976,9 +976,6 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
 //                     C container, int index,  // Arguments for default implementation
 //                     StoreVectorMaskedOperation<C, V, M> defaultImpl) {
 //
-//    TODO: Handle special cases where load/store happens from/to byte array but element type
-//    is not byte.
-//
 bool LibraryCallKit::inline_vector_mem_masked_operation(bool is_store) {
   const TypeInstPtr* vector_klass = gvn().type(argument(0))->isa_instptr();
   const TypeInstPtr* mask_klass   = gvn().type(argument(1))->isa_instptr();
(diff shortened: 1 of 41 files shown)


XiaohongGong:
Hi @PaulSandoz @jatin-bhateja , could you please do me a favor to kindly take a look at this PR? Thanks in advance!

PaulSandoz (commented):
I browsed through but need to go through in more detail, but i have a question. You fixed an issue with `AbstractMask.checkIndexByLane`. I am wondering if our tests are insufficient or that this change exposed a bug and the tests previously passed because the fallback correctly failed when accessing out-of-bounds?

XiaohongGong:
> I browsed through but need to go through in more detail, but i have a question. You fixed an issue with `AbstractMask.checkIndexByLane`. I am wondering if our tests are insufficient or that this change exposed a bug and the tests previously passed because the fallback correctly failed when accessing out-of-bounds?

Yes, I think this change exposed this issue and the tests previously passed because the default java implementation could throw the expected IOOB exception. The original masked ` intoByteArray/intoByteBuffer` APIs are not vectorized.

PaulSandoz (approved):
Java changes look good.

XiaohongGong:
Thanks for the review @PaulSandoz !

XiaohongGong:
@jatin-bhateja , could you please help to take a look at this patch especially the compiler part? Thanks so much!

XiaohongGong:
Hi @jatin-bhateja, I have to integrate this PR now in case it blocks any further development. Any comments are still welcome and we can revisit them in future if needed. Thanks so much!

PR #108, opened 2021-08-06
8272100: VectorAPI: modify existing implementation of masked neg and not operation.

jatin-bhateja:
- Masked NEG/NOT operations can be implemented using existing masked operations instead of emitting an explicit vector blend after the operation. 
- Changes shows around 1.3-1.5X gain when merged with X86 backend changes for masking operations. https://github.com/openjdk/panama-vector/pull/99

---------
### Progress

### Issue
 * [JDK-8272100](https://bugs.openjdk.java.net/browse/JDK-8272100): VectorAPI: modify existing implementation of masked neg and not operation.

### Reviewers
 * [Xiaohong Gong](https://openjdk.java.net/census#xgong) (@XiaohongGong - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/108/head:pull/108` \
`$ git checkout pull/108`

`$ git checkout pull/108` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/108/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 108`

`$ git pr show -t 108`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/108.diff">https://git.openjdk.java.net/panama-vector/pull/108.diff</a>

</details>


src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -605,8 +605,10 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
                 return blend(broadcast(-1), compare(NE, 0, m));
             }
 #if[BITWISE]
-            if (op == NOT || op == NEG) {
-                return blend(lanewise(op), m);
+            if (op == NOT) {
+                return lanewise(XOR, broadcast(-1), m);
+            } else if (op == NEG) {
+                return lanewise(NOT, m).lanewise(ADD, broadcast(1), m);
```
PaulSandoz, line 611: Why not stick to the same pattern is the non-masked `lanewiseTemplate` method?
```suggestion
            if (op == NOT) {
                return broadcast(-1).lanewise(XOR, this, m);
            } else if (op == NEG) {
                return broadcast(0).lanewise(SUB, this, m);
```

XiaohongGong: Seems the result is not equal to `this` vector for the non-active lanes (false lane for `m`)?

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
```diff
@@ -590,8 +590,10 @@ ByteVector lanewiseTemplate(VectorOperators.Unary op,
             if (op == ZOMO) {
                 return blend(broadcast(-1), compare(NE, 0, m));
             }
-            if (op == NOT || op == NEG) {
-                return blend(lanewise(op), m);
+            if (op == NOT) {
+                return lanewise(XOR, broadcast(-1), m);
+            } else if (op == NEG) {
+                return lanewise(NOT, m).lanewise(ADD, broadcast(1), m);
```
XiaohongGong, line 596: Why not directly intrinsify `NEG` as an unary op and generate the "NegVI" node in compiler? SVE supports the masked neg instruction. How about x86 AVX-512?

jatin-bhateja: For targets which do not support direct vector NEG, logic will need to dismantled into vector NOT + ADD.
Other approach was to emit vector multiplication with -1. But it will be costly for sub-word types (byte/word). I think we can take any approach, keep this java change as its and add compiler transform which folds the logic into NegVI during XOR idealization since it will be specific for one target (SVE) currently.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
jatin-bhateja: Yes, result will not be same since broadcasted zero vector is first operand. We need to select the lanes of 'this' vector when corresponding mask bit is false.

PaulSandoz: Doh! Can you place a comment the NOT/ADD expression? e.g.
```
// Cannot perform broadcast(0).lanewise(SUB, this, m)
// since we need to retain unselected lanes in this vector  
```

jatin-bhateja:
@reviwers, kindly let me know if its good to check this in.

XiaohongGong (approved):
LGTM, thanks!

PR #115, opened 2021-08-27
8273057: [vector] New VectorAPI "SelectiveStore"

JoshuaZhuwj:
Hi,

I want to propose a new VectorAPI "Selective Store/Load" and share my
implementation. Currently Alibaba's internal databases are in the
process of applying VectorAPI and they have requirements on "Selective
Store" for acceleration.

My proposed VectorAPI is declared as below [1]:
```
    int selectiveIntoArray($type$[] a, int offset, VectorMask<$Boxtype$> m);
```
The active elements (with their respective bit set in mask) are
contiguously stored into the array "a". Assume N is the true count of
mask, the elements starting from a[offset+N] till a[offset+laneCount]
are left unchanged. The return value represents the number of elements
store into the array and "offset + return value" is the new offset of
the next iteration.
image
This API will be used like the following manner [2]:
```
    tld.conflict_cnt = 0;
    for (int i = 0; i < ARRAY_LENGTH; i += INT_PREFERRED_SPECIES.length()) {
      IntVector av = IntVector.fromArray(INT_PREFERRED_SPECIES, tld.int_input1, i);
      IntVector bv = IntVector.fromArray(INT_PREFERRED_SPECIES, tld.int_input2, i);
      IntVector cv = IntVector.fromArray(INT_PREFERRED_SPECIES, tld.int_index, i);
      VectorMask<Integer> mask = av.compare(VectorOperators.NE, bv);
      tld.conflict_cnt += cv.selectiveIntoArray(tld.conflict_array, tld.conflict_cnt, mask);
    }
```
My patch includes the following changes:
&nbsp;&nbsp;* Selective Store VectorAPI for Long & Int
&nbsp;&nbsp;* Assembler: add x86 instruction "VPCOMPRESSD" and "VPCOMPRESSQ"
&nbsp;&nbsp;* Instruction selection: vselective_store; kmask_truecount (true count of kregister)
&nbsp;&nbsp;* Add node "StoreVectorSelective"
&nbsp;&nbsp;* Add a new parameter "is_selective" in inline_vector_mem_masked_operation()
&nbsp;&nbsp;&nbsp;&nbsp;in order to distinguish Masked version or Selective version
&nbsp;&nbsp;* jtreg cases
&nbsp;&nbsp;* JMH benchmark
      
TODO parts I will implement:
&nbsp;&nbsp;* Selective Store for other types
&nbsp;&nbsp;* Selective Load
&nbsp;&nbsp;* Some potential optimization. Such as: when mask is allTrue, SelectiveIntoArray() -> IntoArray()

Test:
&nbsp;&nbsp;* Passed VectorAPI jtreg cases.
&nbsp;&nbsp;* Result of JMH benchmark to evaluate API's performance in Alibaba's real scenario.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;UseAVX=3; thread number = 8; conflict data percentage: 20% (that means 20% of mask bits are true)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;http://cr.openjdk.java.net/~jzhu/8273057/jmh_benchmark_result.pdf

[1] https://github.com/JoshuaZhuwj/panama-vector/commit/69623f7d6a1eae532576359328b96162d8e16837#diff-13cc2d6ec18e487ddae05cda671bdb6bb7ffd42ff7bc51a2e00c8c5e622bd55dR4667
[2] https://github.com/JoshuaZhuwj/panama-vector/commit/69623f7d6a1eae532576359328b96162d8e16837#diff-951d02bd72a931ac34bc85d1d4e656a14f8943e143fc9282b36b9c76c1893c0cR144
[3] failed to inline (intrinsic) by https://github.com/openjdk/panama-vector/blob/60aa8ca6dc0b3f1a3ee517db167f9660012858cd/src/hotspot/cpu/x86/x86.ad#L1769

Best Regards,
Joshua

---------
### Progress

### Issue
 * [JDK-8273057](https://bugs.openjdk.java.net/browse/JDK-8273057): [vector] New VectorAPI "SelectiveStore"

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/115/head:pull/115` \
`$ git checkout pull/115`

`$ git checkout pull/115` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/115/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 115`

`$ git pr show -t 115`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/115.diff">https://git.openjdk.java.net/panama-vector/pull/115.diff</a>

</details>


JoshuaZhuwj:
Hi Paul,

Thanks a lot for your quick reply and share of design thinking.

> We already have general mask accepting scatter/gather store/load. (I have always been a bit uncertain whether we have the right signatures for these methods, and whether they are necessary if we can use shuffles.)
> 
> To use the scatter store method today for your use-case we would have to:
> 
> - compute an int[] array from the set lanes of the mask, M say
> - computing the ?prefix" mask from the set lanes of M, PM
> - store the vector using the int[] array and PM.
> 
> Another alternative is to:
> 
> - compute a ?compression? shuffle, S, from the set lanes of the mask, M
> - apply S to the vector, produce a compressed vector CV
> - computing the ?prefix" mask from the set lanes of M, PM
> - store CV using M

In my opinion, no matter the first scatter store way or the second shuffle way,
the key to resolve this use case is similar: how to generate "compression" shuffle or "compression" index array according to the set lanes of the mask.
(The "prefix" mask could be computed by true count of mask.)
But when compose existing primitives, how much performance is affected is also important compared to native instruction support or scalar version.

When I help database developers to apply VectorAPI, they usually care:
&nbsp;&nbsp;*how much performance they could gain;
&nbsp;&nbsp;*VectorAPI's ease of use to reduce code complexity and ensure portability;
&nbsp;&nbsp;*whether intrinsics are inlined in their scenarios (fail cases bring performance degradation).

Four vectorized data movement primitives: selective load, selective store, gather, and scatter
are critical SIMD operations for in-memory databases.
Currently gather and scatter are supported in VectorAPI.
That's why I help them implement selective ops and try to propose "Selective Store/Load" in VectorAPI.

> The primitive I am searching for might be a way to create a shuffle from a mask.

What I know for creating shuffle from mask is to leverage a precomputed, cache resident table.
Take four bits mask as example,
```
mask_index | mask |  index_array
    0      | 0000 |  
    1      | 0001 |  3
    2      | 0010 |  2
    3      | 0011 |  2, 3
    4      | 0100 |  1
    5      | 0101 |  1, 3
   ...
    14     | 1110 |  0, 1, 2
    15     | 1111 |  0, 1, 2, 3
```
The index_array could be treated as index map or converted into shuffle.

> Let?s say we could write:
> 
> Int[] a = ...
> IntVector v = ...
> VectorMask m = ?
> 
> // The new primitive, create a shuffle from the mask that partitions vector elements
> // according to the set and unset mask lane elements.
> VectorShuffle s = m.toPartitioningShuffle();
> // Partition the elements
> IntVector cv = v.rearrange(s);
> 
> // This method is likely not optimal, yet!
> // Another method could be added that is prefix(int length)
> VectorMask pm = m.species().indexInRange(0, m.trueCount());
> 
> // Use existing masked store
> cv.intoArray(a, offset, pm);
> // Increase offset by number of stores
> offset += m.trueCount();
> 
> Is it possible for C2 to detect the kind of shuffle pattern and masking to sufficiently optimize? Experts please chime in!

I think the pattern will be too complicated and the optimization may be not reliable.

Thanks,
Joshua

JoshuaZhuwj:
> I think I would rather see a vector-to-vector compress operation, than a vector-to-memory operation that also includes compression. Isn?t that the real underlying primitive?

Agree. John, thanks a lot for your review comments. This will make the primitive more friendly.

nsjian:
> > I think I would rather see a vector-to-vector compress operation, than a vector-to-memory operation that also includes compression. Isn?t that the real underlying primitive?
> 
> Agree. John, thanks a lot for your review comments. This will make the primitive more friendly.

Yes, if we have to introduce a new API, a vector-to-vector compress operation sounds more reasonable. Arm SVE instrctuion COMPACT could also do such work.

JoshuaZhuwj:
Thank Paul, John Rose and Ningsheng for your comments.

A vector-to-vector compress operation is the more friendly primitive.
Both Intel AVX3 instruction "COMPRESS" and Arm SVE instruction "COMPACT" provide the capability.
Hence selective store could be implemented by
```
    VectorMask<Integer> mask = ...;
    IntVector bv = av.compress(mask);
    VectorMask<Integer> prefixMask = prefix(mask.trueCount());
    bv.intoArray(array, offset, prefixMask);
    offset += mask.trueCount();
```
The vector-to-vector compress primitive together with a store and prefix mask could be optimized further into memory destination version of compress on supported architecture.

Once architectures do not support vector-to-vector compress natively,
the intrinsic bailed out and then java path is taken.
Java path could be composed by "mask -> shuffle -> rearrange" and C2 will then try to inline intrinsics again for these API calls.
In this way, we are also able to implement scatter/gather store/load in the same composition manner to keep design consistency.

Best Regards,
Joshua

JoshuaZhuwj:
> In summary, from the discussion, compress/expand are:
> 
> - important conceptually, even if the same functionality could be composed from shuffles (such as used by an implementation); and
> 
> - at the right level to reliably optimize on supporting hardware.
> 
> So specification-wise we introduce expanding/compressing cross-lane operations. API-wise I prefer two distinct methods rather that one that accepts boolean indicating expansion or compression. We can declare one intrinsic method in VectorSupport.

Thanks Paul for bringing the deep design thinking.
John rose's encompassing knowledge impresses me a lot.
Also thank Sandhya and Ningsheng for comments.

I will refactor my codes and implement these two cross-lane data movement primitives: mask-based compression & expansion.
They will work on general vectors and be declared:
```
    $abstractvectortype$ compress(VectorMask<$Boxtype$> m);
    $abstractvectortype$ expand(VectorMask<$Boxtype$> m);
```
These two vector-to-vector operations together with a store/load and prefix mask could be optimized further into single memory version instruction on supported architecture.

jatin-bhateja:
> > In summary, from the discussion, compress/expand are:
> > 
> > * important conceptually, even if the same functionality could be composed from shuffles (such as used by an implementation); and
> > * at the right level to reliably optimize on supporting hardware.
> > 
> > So specification-wise we introduce expanding/compressing cross-lane operations. API-wise I prefer two distinct methods rather that one that accepts boolean indicating expansion or compression. We can declare one intrinsic method in VectorSupport.
> 
> Thanks Paul for bringing the deep design thinking.
> John Rose's encompassing knowledge impresses me a lot.
> Also thank Sandhya and Ningsheng for comments.
> 
> I will refactor my codes and implement these two cross-lane data movement primitives: mask-based compression & expansion.
> They will work on general vectors and be declared:
> 
> ```
>     $abstractvectortype$ compress(VectorMask<$Boxtype$> m);
>     $abstractvectortype$ expand(VectorMask<$Boxtype$> m);
> ```
> 
> These two vector-to-vector operations together with a store/load and prefix mask could be optimized further into single memory version instruction on supported architecture.

Hi @JoshuaZhuwj , 

While adding new macro level APIs is appealing, we can also extend following existing vectorAPIs to accept another boolean flag "is_selective" under which compression/expansion triggers. In this use case its difficult to infer COMPRESSION through Auto-vectorizer though we made attempts in past to infer complex loop patterns for VNNI instruction. 

 ```
public static IntVector fromArray​(VectorSpecies<Integer> species,
 int[] a,
 int offset,
 VectorMask<Integer> m) 


public final void intoArray​(int[] a,
 int offset,
 VectorMask<Integer> m)

```
This way we can also share common optimizations as you suggested earlier to convert masked COMPRESS to unmasked vector move for ALLTRUE mask, some work[1][2] is already in place on this front. 

Best Regards,
Jatin

[1] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L752
[2] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L771

**Comment by JoshuaZhuwj** [comment]
> While adding new macro level APIs is appealing, we can also extend following existing vectorAPIs to accept another boolean flag "is_selective" under which compression/expansion triggers.
> 
> ```
> public static IntVector fromArray​(VectorSpecies<Integer> species,
> int[] a,
> int offset,
> VectorMask<Integer> m) 
> 
> 
> public final void intoArray​(int[] a,
> int offset,
> VectorMask<Integer> m)
> ```

Per design discussion in this thread, compared to vector-to-memory operation, vector-to-vector compress/expand operation is the more friendly primitive.
It can also be used to "bridge to and from permutation simply by working with index vectors like iota, and perhaps (as sugar) lifting selected vector operations to shuffles."
For different architectures, like SVE, memory destination version is also not supported natively.

> In this use case its difficult to infer COMPRESSION through Auto-vectorizer though we made attempts in past to infer complex loop patterns for VNNI instruction.

Could you elaborate on it please? I do not follow this.

> This way we can also share common optimizations as you suggested earlier to convert masked COMPRESS to unmasked vector move for ALLTRUE mask, some work[1][2] is already in place on this front.
> 
> [1] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L752
> [2] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L771

Yes. Since compress/expand op is also mask-based, this piece of optimization is common. Maybe we can think of one way to share this optimization for different kinds of masked operations?

**Comment by jatin-bhateja** [comment]
> > While adding new macro level APIs is appealing, we can also extend following existing vectorAPIs to accept another boolean flag "is_selective" under which compression/expansion triggers.
> > ```
> > public static IntVector fromArray​(VectorSpecies<Integer> species,
> > int[] a,
> > int offset,
> > VectorMask<Integer> m) 
> > 
> > 
> > public final void intoArray​(int[] a,
> > int offset,
> > VectorMask<Integer> m)
> > ```
> 
> Per design discussion in this thread, compared to vector-to-memory operation, vector-to-vector compress/expand operation is the more friendly primitive.
> It can also be used to "bridge to and from permutation simply by working with index vectors like iota, and perhaps (as sugar) lifting selected vector operations to shuffles."
> For different architectures, like SVE, memory destination version is also not supported natively.
> 
> > In this use case its difficult to infer COMPRESSION through Auto-vectorizer though we made attempts in past to infer complex loop patterns for VNNI instruction.
> 
> Could you elaborate on it please? I do not follow this.
>

I meant auto-vectorizing following loop which mimic compression semantics could be tricky if its difficult to ascertain the independence between memory references. Like in following case 'j' could be a middle index in array and thus if the distance between memory references is less than chosen vector width it may result into incorrect overwrite. 
  
```
for ( int i = 0; i < n ; i++ ) {
    if ( mask[i] > 1 ) {
         a[j++] = a[i];
     }
}
```

> > This way we can also share common optimizations as you suggested earlier to convert masked COMPRESS to unmasked vector move for ALLTRUE mask, some work[1][2] is already in place on this front.
> > [1] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L752
> > [2] https://github.com/openjdk/panama-vector/blob/master/src/hotspot/share/opto/vectornode.cpp#L771
> 
> Yes. Since compress/expand op is also mask-based, this piece of optimization is common. Maybe we can think of one way to share this optimization for different kinds of masked operations?

Yes, agree.

JoshuaZhuwj:
> Would it be possible for you to target this work to a new branch? e.g. vectorIntrinsics+compress.
> 
> I want to avoid complicating review and integration of the masking work for JEP 417. We are close to merging vectorIntrinsics+mask into vectorIntrinsics and then starting reviews of the masking support.

Yes, of course. Thank you, Paul. Hence could you please help me checkout a new branch based on "vectorIntrinsics+mask" or "vectorIntrinsics" after merging?

> It may be possible that compress/expand become part of JEP 417, if we can get it ready in time. If so I can update the JEP accordingly and help do CSRs. Otherwise, we can target to the next round of incubation.

I'm fine with the plan to place compress/expand in the next round of incubation. I'm afraid I cannot catch the last bus before the review of masking support while I have other works at hand.

JoshuaZhuwj:
> > > In this use case its difficult to infer COMPRESSION through Auto-vectorizer though we made attempts in past to infer complex loop patterns for VNNI instruction.
> > 
> > Could you elaborate on it please? I do not follow this.
> 
> I meant auto-vectorizing following loop which mimic compression semantics could be tricky if its difficult to ascertain the independence between memory references. Like in following case 'j' could be a middle index in array and thus if the distance between memory references is less than chosen vector width it may result into incorrect overwrite.
> 
> ```
> for ( int i = 0; i < n ; i++ ) {
>     if ( mask[i] > 1 ) {
>          a[j++] = a[i];
>      }
> }
> ```

Agree. I think current SWPointer compare (used to detect possible same address) does not apply to this pattern.
In this case, besides SWPointer comparison, "j" cannot be inferred by induction varaible "i" is more critical for auto-vectorization.

Do you have plan to support masked operations in SLP optimization?

jatin-bhateja:
> > > > In this use case its difficult to infer COMPRESSION through Auto-vectorizer though we made attempts in past to infer complex loop patterns for VNNI instruction.
> > > 
> > > 
> > > Could you elaborate on it please? I do not follow this.
> > 
> > 
> > I meant auto-vectorizing following loop which mimic compression semantics could be tricky if its difficult to ascertain the independence between memory references. Like in following case 'j' could be a middle index in array and thus if the distance between memory references is less than chosen vector width it may result into incorrect overwrite.
> > ```
> > for ( int i = 0; i < n ; i++ ) {
> >     if ( mask[i] > 1 ) {
> >          a[j++] = a[i];
> >      }
> > }
> > ```
> 
> Agree. I think current SWPointer compare (used to detect possible same address) does not apply to this pattern.
> In this case, besides SWPointer comparison, "j" cannot be inferred by induction varaible "i" is more critical for auto-vectorization.

A PhiNode will be created for j for sure, and it's indeed an induction variable since with each iteration its value is getting incremented.  Yes, its not a secondary induction variable in this case and which makes any dependency analysis between array references indexed by different subscripts 'i' or 'j' difficult. 

> 
> Do you have plan to support masked operations in SLP optimization?

Given that SLP algorithm does not works across basic blocks a simplified approach will be to add ideal transformation post SLP to replace VectorBlend + VectorOperation to a masked operation where ever applicable

JoshuaZhuwj:
> I started off the discussion last week as part of https://github.com/[/pull/143](https://github.com/openjdk/panama-vector/pull/143). The email thread, with additional inputs, is at: https://mail.openjdk.java.net/pipermail/panama-dev/2021-October/015223.html
> 
> Once you are back from vacation, please do join the discussion and development.

Okay. I will send out what I already implemented to avoid possible repetitive work.

PR #126, opened 2021-09-17
8273949: Intrinsic creation for VectorMask.toLong operation.

jatin-bhateja:
Summary of changes:
- Intrinsification of VectorMask.toLong() API.
- Supports inline expansion for both AVX512 and non-AVX512 targets.
- Used toLong() API to optimize existing Java API implementation of VectorMask.laneIsSet() operation.

Following performance number are generated using JMH benchmark modification included with the patch.

System:  Intel(R) Xeon(R) Platinum 8280 CPU @ 2.70GHz (Cascade Lake Server 28C 2S)

Benchmark | VECSIZE | Baseline Score (ops/ms) | With Opt (ops/ms) | Gain Ratio
-- | -- | -- | -- | --
MaskQueryOperationsBenchmark.testToLongByte | 128 | 90451.424 | 346941.379 | 3.835665196
MaskQueryOperationsBenchmark.testToLongByte | 256 | 63127.764 | 338331.425 | 5.359471072
MaskQueryOperationsBenchmark.testToLongByte | 512 | 40543.264 | 313836.333 | 7.740776199
MaskQueryOperationsBenchmark.testToLongLong | 128 | 171989.714 | 152872.758 | 0.88884826
MaskQueryOperationsBenchmark.testToLongLong | 256 | 164702.273 | 324794.578 | 1.972010295
MaskQueryOperationsBenchmark.testToLongLong | 512 | 122667.916 | 318060.096 | 2.59285481
MaskQueryOperationsBenchmark.testToLongShort | 128 | 122656.408 | 346691.082 | 2.826522378
MaskQueryOperationsBenchmark.testToLongShort | 256 | 96838.555 | 360909.28 | 3.726917239
MaskQueryOperationsBenchmark.testToLongShort | 512 | 63119.009 | 313075.159 | 4.960077225
MaskQueryOperationsBenchmark.testToLonglong | 128 | 180855.623 | 324620.433 | 1.794914792
MaskQueryOperationsBenchmark.testToLonglong | 256 | 122705.631 | 324315.916 | 2.643040204
MaskQueryOperationsBenchmark.testToLonglong | 512 | 90396.687 | 324318.095 | 3.58772103

---------
### Progress

### Issue
 * [JDK-8273949](https://bugs.openjdk.java.net/browse/JDK-8273949): Intrinsic creation for VectorMask.toLong operation.

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 6b38d057e4ad9908714a2f45d2e99c7c59231e2c
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to f1816ce81c05729b51118df7a4f97553a4eb202c
 * [Eric Liu](https://openjdk.java.net/census#eliu) (@theRealELiu - Committer) ⚠️ Review applies to f1816ce81c05729b51118df7a4f97553a4eb202c

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/126/head:pull/126` \
`$ git checkout pull/126`

`$ git checkout pull/126` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/126/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 126`

`$ git pr show -t 126`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/126.diff">https://git.openjdk.java.net/panama-vector/pull/126.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -653,7 +653,7 @@ bool LibraryCallKit::inline_vector_shuffle_iota() {
 }
 
 // <E, M>
-// int maskReductionCoerced(int oper, Class<? extends M> maskClass, Class<?> elemClass,
+// long maskReductionCoerced(int oper, Class<? extends M> maskClass, Class<?> elemClass,
 //                          int length, M m, VectorMaskOp<M> defaultImpl)
 bool LibraryCallKit::inline_vector_mask_operation() {
   const TypeInt*     oper       = gvn().type(argument(0))->isa_int();


src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
```diff
@@ -63,23 +63,11 @@
...
-        if (length() > Long.SIZE) {
-            throw new UnsupportedOperationException("too many lanes for one long");
-        }
-        long res = 0;
-        long set = 1;
-        boolean[] bits = getBits();
-        for (int i = 0; i < bits.length; i++) {
-            res = bits[i] ? res | set : res;
-            set = set << 1;
+        if (length() <= Long.SIZE) {
```
PaulSandoz, line 66: We need to check that `i > 0 && i < length()` and if not throw `IllegalArgumentException`.

Currently the behavior is unspecified, can you please add the following to the documentation of `VectorMask.laneIsSet`. (I will handle any CSR related tasks.)

```
     * @throws IllegalArgumentException if the index is is out of range
     * ({@code < 0 || >= length()})
```

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
```diff
@@ -708,22 +708,32 @@ Byte128Mask xor(VectorMask<Byte> mask) {
         @Override
         @ForceInline
         public int trueCount() {
-            return VectorSupport.maskReductionCoerced(VECTOR_OP_MASK_TRUECOUNT, Byte128Mask.class, byte.class, VLENGTH, this,
-                                                      (m) -> trueCountHelper(((Byte128Mask)m).getBits()));
+            return (int) VectorSupport.maskReductionCoerced(VECTOR_OP_MASK_TRUECOUNT, Byte128Mask.class, byte.class, VLENGTH, this,
+                                                      (m) -> (long)trueCountHelper(((Byte128Mask)m).getBits()));
```
PaulSandoz, line 712:
```suggestion
            return (int) VectorSupport.maskReductionCoerced(VECTOR_OP_MASK_TRUECOUNT, Byte128Mask.class, byte.class, VLENGTH, this,
                                                      (m) -> trueCountHelper(m.getBits()));
```

We don't need the cast to `long` for the return from the lambda expression. And I think the cast of the mask is redundant too. Same applies to the other three methods.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
```diff
@@ -63,6 +63,9 @@
 
     @Override
     public boolean laneIsSet(int i) {
+        if (i < 0 || i >= length()) {
+            throw new IllegalArgumentException("Index " + i + " must be zero or positive, and less than " + length());
+        }
```
PaulSandoz: Apologies i should have previously said use `Objects.checkIndex(i, length())`. In fact lets call `length` once:
```
    int length = length();
    Objects.checkIndex(i, length);
    if (length <= Long.SIZE) {
    ...
```

Separately i wonder if the second bounds check should be `if i <= Long.SIZE`, since the first bounds check should dominate in many cases?

(I know there are other areas where we should `Objects.checkIndex` e.g. `*MaxVector.lane/withLane` but we can fix those later.)

jatin-bhateja: With length <= Long.SIZE there is a better possibility of optimizing out else part and thus only fast path gets generated. length is final static values and hence in most of the cases due to in-lining this constant may get propagated.

PaulSandoz (approved):
Java changes are good.

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -957,6 +957,14 @@ class VectorMaskLastTrueNode : public VectorMaskOpNode {
   virtual int Opcode() const;
 };
 
+class VectorMaskToLongNode : public VectorMaskOpNode {
+ public:
+  VectorMaskToLongNode(Node* mask, const Type* ty):
+    VectorMaskOpNode(mask, ty, Op_VectorMaskLastTrue) {}
```
sviswa7, line 963: Shouldn't this be Op_VectorMaskToLong?

src/hotspot/share/prims/vectorSupport.cpp
```diff
@@ -430,6 +430,18 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
...
     }
+    case VECTOR_OP_MASK_TOLONG: {
+      switch (bt) {
+        case T_BYTE:  // fall-through
+        case T_SHORT: // fall-through
+        case T_INT:   // fall-through
+        case T_LONG:  // fall-through
+        case T_FLOAT: // fall-through
+        case T_DOUBLE: return Op_VectorMaskToLong;
+        default: fatal("MASK_TRUECOUNT: %s", type2name(bt));
```
sviswa7, line 441: MASK_TOLONG here? Also break is missing.

jatin-bhateja:
@nsjian , kindly check the IR change, it may need some change in AARCH64 backend since now VectorStoreMask is not being inserted while connecting mask generating node to mask operation node. This saves redundant store mask operation if target supports predicate registers.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8659,42 +8660,56 @@ instruct vmasked_store64(memory mem, vec src, kReg mask) %{
   ins_pipe( pipe_slow );
 %}
 
-instruct vmask_truecount_evex(rRegI dst, vec mask, rRegL tmp, kReg ktmp, vec xtmp) %{
-  predicate(VM_Version::supports_avx512vlbw());
-  match(Set dst (VectorMaskTrueCount mask));
-  effect(TEMP_DEF dst, TEMP tmp, TEMP ktmp, TEMP xtmp);
-  format %{ "vector_truecount_evex $mask \t! vector mask true count" %}
+instruct vmask_tolong_evex(rRegL dst, kReg mask) %{
```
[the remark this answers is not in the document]
jatin-bhateja: Its already covered under _LP64

src/hotspot/share/prims/vectorSupport.cpp
jatin-bhateja: This is consistent with other case statements.

nsjian:
> @nsjian , kindly check the IR change, it may need some change in AARCH64 backend since now VectorStoreMask is not being inserted while connecting mask generating node to mask operation node. This saves redundant store mask operation if target supports predicate registers.

Hi @jatin-bhateja , thanks for the heads-up. The VectorMaskOp connection change looks reasonable. We will update SVE backend later.

src/hotspot/share/prims/vectorSupport.cpp
sviswa7: default: fatal("MASK_TRUECOUNT: %s", type2name(bt));
should refer to MASK_TOLONG.

sviswa7 (approved):
Only one comment on cut and paste mistake in error message, Please fix that. Other than that the patch looks good. No need for re-rereview.

theRealELiu (approved):
LGTM. 

Just the comment is mismatched.
``` java
// <E, M>
// long maskReductionCoerced(int oper, Class<? extends M> maskClass, Class<?> elemClass,
//                          int length, M m, VectorMaskOp<M> defaultImpl)
bool LibraryCallKit::inline_vector_mask_operation() {
```

DamonFool:
Build failure on MacOS after this PR: https://github.com/openjdk/panama-vector/pull/133 .

DamonFool:
Incorrect computation after this pr.
Please see: https://github.com/openjdk/panama-vector/pull/138 .
Thanks.

PR #128, opened 2021-09-18
(Draft) Optimize long range checks

rsmogura:
I wonder if you would have time to pre-check this. I know that elimination of range check is quite important
and if this is a good approach.

Shortly I tried to extend current RCE by adding support for long and mitigate possible overflows over long
values with `Overflow` nodes - creating predicate if value is less than rang and does not overflow.

It's not complete.

---

Try to optimize long RCE by adding loop predicates, for loops using
Long RC (`CmpUL`).

This code is intended to handle int counted loops with Long RC, so
as well striped loops.
* modify `is_scaled_iv_plus_offset` to allow searching for induction
  variables over _long_ nodes
* modify `rc_predicate` to add overflow checks if long could overflow
* and bail-out conditions to  `PhaseIdealLoop::split_thru_phi` to prevent
   this optimizations (this code would not be optimized otherwise
   `for (long l=2; l < length; l += 4) Objects.checkIndex(2 + l, length);`
* prevent long RCE when scale is larger than `1 << 16`,
   for now there's not overflow check code;
This is inital commit, just to validate if this is a good approach, and
checks are ok.

Especially worth for double checks are some assumptions about overflow.

Later we could:
- better handle SubL, scale, and multiplications
in `is_scaled_iv_plus_offset`
- create extended Add, Mul, Sub projecting value and overflow
(or better all CR flags) - will increase performance slightly,
- better optimize Overflow nodes `Ideal` in IGVN

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-vector.git pull/128/head:pull/128` \
`$ git checkout pull/128`

`$ git checkout pull/128` \
`$ git pull https://git.openjdk.org/panama-vector.git pull/128/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 128`

`$ git pr show -t 128`

</details>

<a href="https://git.openjdk.org/panama-vector/pull/128.diff">https://git.openjdk.org/panama-vector/pull/128.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/c2_globals.hpp
+++ b/src/hotspot/share/opto/c2_globals.hpp
@@ -767,7 +767,10 @@
                                                                             \
   product(bool, VerifyReceiverTypes, trueInDebug, DIAGNOSTIC,               \
           "Verify receiver types at runtime")                               \
-
+                                                                            \
+  product(bool, LongRCE, true, DIAGNOSTIC,                                  \
+          "Enable elimination of long range checks in int"                  \
+          " (& stripped) loops")                                            \
 // end of C2_FLAGS
 
 DECLARE_FLAGS(C2_FLAGS)
--- a/src/hotspot/share/opto/loopPredicate.cpp
+++ b/src/hotspot/share/opto/loopPredicate.cpp
@@ -30,6 +30,8 @@
 #include "opto/convertnode.hpp"
 #include "opto/loopnode.hpp"
 #include "opto/matcher.hpp"
+#include "opto/mathexactnode.hpp"
+#include "opto/movenode.hpp"
 #include "opto/mulnode.hpp"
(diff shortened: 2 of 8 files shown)


PaulSandoz:
@rsmogura (i realize you are using panama-vector as the repo for experimentation here) Are you aware of https://github.com/openjdk/jdk/pull/2045 ? @rwestrel may have some comments to your approach and perhaps whether you can help on 2045?

rsmogura:
Thanks @PaulSandoz I'll ask @rwestrel - maybe this can be some useful input.

PR #141, opened 2021-10-01
8274631: assert(false) failed: bad AD file with -XX:UseKNLSetting

jatin-bhateja:
- Fix for assertion failure in test/hotspot/jtreg/compiler/vectorapi/VectorMaskCastTest.java crashes during compilation over KNL target.
- A safety check has been added to prevent intrinsification of VectorMask.cast API if source and destination have incompatible ideal types i.e one is predicate vector type (TypeVectMask) and other a vector type (TypeVect).

---------
### Progress

### Issue
 * [JDK-8274631](https://bugs.openjdk.java.net/browse/JDK-8274631): assert(false) failed: bad AD file with -XX:UseKNLSetting

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/141/head:pull/141` \
`$ git checkout pull/141`

`$ git checkout pull/141` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/141/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 141`

`$ git pr show -t 141`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/141.diff">https://git.openjdk.java.net/panama-vector/pull/141.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -2410,6 +2410,9 @@ bool LibraryCallKit::inline_vector_convert() {
 
   // Safety check to prevent casting if source mask is of type vector
   // and destination mask of type predicate vector and vice-versa.
+  // From X86 standpoint, this case will only arise over KNL target,
+  // where certain masks (depending on the species) are either propagated
+  // through a vector or predicate register.
   if (is_mask &&
       ((src_type->isa_vect() && dst_type->isa_vectmask()) ||
       (dst_type->isa_vect() && src_type->isa_vectmask()))) {


PaulSandoz:
Verified test now passes with fix.

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -2408,6 +2408,14 @@ bool LibraryCallKit::inline_vector_convert() {
   const TypeVect* src_type = TypeVect::make(elem_bt_from, num_elem_from, is_mask);
   const TypeVect* dst_type = TypeVect::make(elem_bt_to, num_elem_to, is_mask);
 
+  // Safety check to prevent casting if source mask is of type vector
+  // and destination mask of type predicate vector and vice-versa.
+  if (is_mask &&
+      ((src_type->isa_vect() && dst_type->isa_vectmask()) ||
+      (dst_type->isa_vect() && src_type->isa_vectmask()))) {
+    return false;
```
PaulSandoz, line 2416: Should we add `f (C->print_intrinsics()) {` in this case?

jatin-bhateja: Intrinsification is getting prevented due to an internal implementation constraint of compiler, I think it will not provide any useful information to user in this case.

PaulSandoz: So IIUC it's because the flags create an artificial scenario that does not ordinarily arise. If so can we add a comment regarding the configuration that might result in this code path being taken?

PaulSandoz: Also can we double check if this only arises for debug builds?

jatin-bhateja: From X86 standpoint, this case will only arise over KNL, where certain masks (depending on the species) are either propagated through a vector or predicate register.  This is exactly what the code checks. I did not add a comment in code since its a common IR code.  Over KNL predicate registers are used only if corresponding vector is 512 bit wide.
Thus a casting b/w Int256Mask and Double512Mask is not being handled over KNL currently under the assumption that most mask are used for vector predication and since KNL does not support EVEX instruction over partial vectors (128 and 256 bits) so masks corresponding these sizes will be propagated using vectors and not predicate registers.

jatin-bhateja: > Also can we double check if this only arises for debug builds?

There are not casting patterns between predicate register and regular vectors. So problem will also show up on non-debug builds.

PaulSandoz (approved):
Thanks for the responses.

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -2408,6 +2408,17 @@ bool LibraryCallKit::inline_vector_convert() {
   const TypeVect* src_type = TypeVect::make(elem_bt_from, num_elem_from, is_mask);
   const TypeVect* dst_type = TypeVect::make(elem_bt_to, num_elem_to, is_mask);
 
+  // Safety check to prevent casting if source mask is of type vector
+  // and destination mask of type predicate vector and vice-versa.
+  // From X86 standpoint, this case will only arise over KNL target,
+  // where certain masks (depending on the species) are either propagated
+  // through a vector or predicate register.
+  if (is_mask &&
+      ((src_type->isa_vect() && dst_type->isa_vectmask()) ||
```
nsjian, line 2417: Unfortunately, this will prevent all vectmask cast operations, since currently a vectmask type is also a vect type. @jatin-bhateja

nsjian: I created JDK-8275025 with the simple fix, but I am not able to verify that on x86. Please have a check: https://github.com/openjdk/panama-vector/pull/151

PR #146, opened 2021-10-07
8274889: Intrinsify Vector.expand/compress APIs for X86

jatin-bhateja:
- Common C2 compiler entry point for compress/expand APIs.
- Inline expansion handling and C2 graph creation.
- For X86 current backed support is added for AVX512 which offers direct compress/expand instructions.

---------
### Progress

### Issue
 * [JDK-8274889](https://bugs.openjdk.java.net/browse/JDK-8274889): Intrinsify Vector.expand/compress APIs for X86

### Reviewers
 * [Eric Liu](https://openjdk.java.net/census#eliu) (@theRealELiu - Committer) ⚠️ Review applies to 26cf1652547d24a0dcb495a2b5c9bc9b72abf785

### Contributors
 * Jatin Bhateja `<jbhateja@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/146/head:pull/146` \
`$ git checkout pull/146`

`$ git checkout pull/146` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/146/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 146`

`$ git pr show -t 146`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/146.diff">https://git.openjdk.java.net/panama-vector/pull/146.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -8742,7 +8742,7 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp, 
 
 // --------------------------------- Compress/Expand Operations ---------------------------
 
-instruct vcompress_expand_reg(vec dst, vec src, kReg mask) %{
+instruct vcompress_expand_reg_evex(vec dst, vec src, kReg mask) %{
   match(Set dst (CompressV src mask));
   match(Set dst (ExpandV src mask));
   format %{ "vector_compress_expand $dst, $src, $mask" %}
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -2707,7 +2707,7 @@ bool LibraryCallKit::inline_vector_compress_expand() {
       vector_klass->const_oop() == NULL || mask_klass->const_oop() == NULL ||
       elem_klass->const_oop() == NULL || !vlen->is_con()) {
     if (C->print_intrinsics()) {
-      tty->print_cr("  ** missing constant: opr=%s vclass=%s mclass =%s etype=%s vlen=%s",
+      tty->print_cr("  ** missing constant: opr=%s vclass=%s mclass=%s etype=%s vlen=%s",
                     NodeClassNames[argument(0)->Opcode()],
                     NodeClassNames[argument(1)->Opcode()],
                     NodeClassNames[argument(2)->Opcode()],
@@ -2750,21 +2750,21 @@ bool LibraryCallKit::inline_vector_compress_expand() {
   Node* opd1 = unbox_vector(argument(5), vbox_type, elem_bt, num_elem);
   if (opd1 == NULL) {
     if (C->print_intrinsics()) {
-      tty->print_cr("  ** unbox failed v2=%s",
+      tty->print_cr("  ** unbox failed vector=%s",
                     NodeClassNames[argument(5)->Opcode()]);
     }
     return false;
   }
 
   ciKlass* mbox_klass = mask_klass->const_oop()->as_instance()->java_lang_Class_klass();
-  assert(is_vector_mask(mbox_klass), "argument(7) should be a mask class");
+  assert(is_vector_mask(mbox_klass), "argument(6) should be a mask class");
   const TypeInstPtr* mbox_type = TypeInstPtr::make_exact(TypePtr::NotNull, mbox_klass);
 
   Node* mask = unbox_vector(argument(6), mbox_type, elem_bt, num_elem);
   if (mask == NULL) {
     if (C->print_intrinsics()) {
       tty->print_cr("  ** unbox failed mask=%s",
-                    NodeClassNames[argument(7)->Opcode()]);
+                    NodeClassNames[argument(6)->Opcode()]);
     }
     return false;
   }
--- a/src/hotspot/share/classfile/vmIntrinsics.hpp
+++ b/src/hotspot/share/classfile/vmIntrinsics.hpp
@@ -1094,7 +1094,8 @@ class methodHandle;
   do_intrinsic(_VectorComExp, jdk_internal_vm_vector_VectorSupport, vector_comexp_op_name, vector_comexp_op_sig, F_S)                           \
    do_signature(vector_comexp_op_sig, "(I"                                                                                                     \
                                       "Ljava/lang/Class;"                                                                                      \
-                                      "Ljava/lang/Class;Ljava/lang/Class;"                                                                     \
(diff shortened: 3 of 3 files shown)


src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -621,6 +625,26 @@ VOUT convert(int oprId,
...
+    }
+
+    @IntrinsicCandidate
+    public static
+    <V extends Vector<E>,
+     M extends VectorMask<E>,
+     E>
+    V comExpOp(int opr,
+              Class<? extends V> vClass, Class<? extends M> mClass, Class<E> eClass,
+              int length, V v1, V v2, M m,
```
JoshuaZhuwj, line 640: No need to use two vectors here. Use zeroing-masking with one vector could meet requirement, right?

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -621,6 +625,26 @@ VOUT convert(int oprId,
...
+                                     M extends VectorMask<?>> {
+        V apply(V v1, V v2, M m);
+    }
+
+    @IntrinsicCandidate
+    public static
+    <V extends Vector<E>,
+     M extends VectorMask<E>,
+     E>
+    V comExpOp(int opr,
```
JoshuaZhuwj, line 638: I prefer boolean instead of int.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8739,6 +8739,22 @@ instruct vmask_first_or_last_true_avx(rRegI dst, vec mask, rRegL tmp, vec xtmp,
...
+
+instruct vcompress_expand_reg(vec dst, vec src, kReg mask) %{
+  match(Set dst (CompressV (Binary dst src) mask));
+  match(Set dst (ExpandV (Binary dst src) mask));
+  format %{ "vector_compress_expand $dst, $src, $mask" %}
+  ins_encode %{
+    int opcode = this->ideal_Opcode();
+    int vector_len = vector_length_encoding(this);
+    BasicType bt  = Matcher::vector_element_basic_type(this);
+    __ compress_expand(opcode, $dst$$XMMRegister, $src$$XMMRegister, $mask$$KRegister, true, bt, vector_len);
```
JoshuaZhuwj, line 8753: Please use zeroing-masking.

src/hotspot/cpu/x86/assembler_x86.hpp
```diff
@@ -2542,6 +2542,22 @@ class Assembler : public AbstractAssembler  {
...
   void vpternlogq(XMMRegister dst, int imm8, XMMRegister src2, XMMRegister src3, int vector_len);
 
+  // Vector compress/expand instructions.
+  void evpcompressb(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpcompressw(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpcompressd(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpcompressq(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evcompressps(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evcompresspd(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void vpcompresspd(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
```
JoshuaZhuwj, line 2552: Please add prefix "e" in the above function name to align with the evex function name convention in this file.

src/hotspot/cpu/x86/assembler_x86.hpp
```diff
@@ -2542,6 +2542,22 @@ class Assembler : public AbstractAssembler  {
...
+
+  void evpexpandb(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpexpandw(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpexpandd(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpexpandq(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
+  void evpexpandps(XMMRegister dst, KRegister mask, XMMRegister src, bool merge, int vector_len);
```
JoshuaZhuwj, line 2558: Please use "evexpandps" instead of "evpexpandps" to match with the instruction name.

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
jatin-bhateja: Two vectors are added in anticipation to support an explicit background vector in future since original APIs had them. 
From the Java side background vector passed currently is ZERO which is why a merging semantics is used in emitted compess/expand instruction instead of zeroing semantics.

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
jatin-bhateja: I have kept the implementation consistent with rest of vector API code and passing an Opcode instead of flag to differentiate between compress/expand operation

jatin-bhateja:
> Jatin, your change will bring around 15% performance degradation for SelectiveStore benchmark using compress VectorAPI.
Hi @JoshuaZhuwj ,

A special check for m.allTrue() has been added which returns original vector argument without any compress/expansion, it favors targets which do not support compress/expand operation.

But I think its better to leave this to user to make this check directly from source instead of making it a part of API implementation since it will effect the performance of fast path.

Will update the patch.

sviswa7:
@jatin-bhateja I have some minor comments. Other than that the patch looks good to me.

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -2734,49 +2732,49 @@ bool LibraryCallKit::inline_vector_compress_expand() {
     return false; // should be primitive type
   }
 
-  const Type* vmask_type = gvn().type(argument(6));
-  if (vmask_type->maybe_null()) {
-    if (C->print_intrinsics()) {
-      tty->print_cr("  ** null mask values are not allowed for compress/expand");
-    }
-    return false;
-  }
```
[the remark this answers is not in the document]
jatin-bhateja: maybe_null() is being checked while unboxing mask argument.

PR #149, opened 2021-10-08
8274975: Add micro benchmark: ALIBABA selective store use case

JoshuaZhuwj:
I separate my implementation of "compress" API into several patches for easy review.
This micro benchmark is extracted from the real use case inside in-memory databases: selectiveStore.
Test scenario: UseAVX=3; thread number = 8; conflict data percentage: 20% (that means 20% of mask bits are true)

---------
### Progress

### Issue
 * [JDK-8274975](https://bugs.openjdk.java.net/browse/JDK-8274975): Add micro benchmark: ALIBABA selective store use case

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to c4d51d6f921b90c3e59949bb67464e0b5949de25

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/149/head:pull/149` \
`$ git checkout pull/149`

`$ git checkout pull/149` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/149/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 149`

`$ git pr show -t 149`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/149.diff">https://git.openjdk.java.net/panama-vector/pull/149.diff</a>

</details>

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/vector/bigdata/SelectiveStore.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/vector/bigdata/SelectiveStore.java
@@ -24,8 +24,8 @@
 
 package org.openjdk.bench.jdk.incubator.vector.bigdata;
 
-import java.util.Random;
 import java.util.concurrent.TimeUnit;
+import java.util.random.RandomGenerator;
 
 import jdk.incubator.vector.IntVector;
 import jdk.incubator.vector.VectorSpecies;
@@ -69,16 +69,16 @@ public class SelectiveStore {
     input2 = new int[ARRAY_LENGTH];
     conflict_array = new int[ARRAY_LENGTH];
 
-    Random rand = new Random();
-    int conflict_id_each_5_elements = rand.nextInt(5);
+    RandomGenerator rng = RandomGenerator.getDefault();
+    int conflict_id_each_5_elements = rng.nextInt(5);
 
     for (int i = 0; i < ARRAY_LENGTH; i++) {
-      index[i] = rand.nextInt(ARRAY_LENGTH);
-      input1[i] = rand.nextInt();
+      index[i] = rng.nextInt(ARRAY_LENGTH);
+      input1[i] = rng.nextInt();
       // Generate 20% conflict data
       int remainder = i % 5;
       if (remainder == conflict_id_each_5_elements) {
-        input2[i] = rand.nextInt();
+        input2[i] = rng.nextInt();
       } else {
         input2[i] = input1[i];
       }


test/micro/org/openjdk/bench/jdk/incubator/vector/bigdata/SelectiveStore.java
```diff
@@ -0,0 +1,189 @@
...
+      VectorMask<Integer> mask = av.compare(VectorOperators.NE, bv);
+      IntVector dv = cv.compress(mask);
+      dv.intoArray(tld.conflict_vector, tld.conflict_vector_cnt, mask.prefixMask());
+      tld.conflict_vector_cnt += mask.trueCount();
+    }
+  }
+
+  @Benchmark
+  @Threads(8)
+  public void selectiveStore_verify(ThreadLocalData tld) {
```
theRealELiu, line 163: Is it necessary to benchmark the correctness?

JoshuaZhuwj: Thanks Eric for your review :-)
This function was imported to ensure compress API's correctness during my development process.
I will remove this function as your suggestion after all dependent patches are merged.

JoshuaZhuwj:
Using vector-to-vector compress, the benchmark performance result:
image

test/micro/org/openjdk/bench/jdk/incubator/vector/bigdata/SelectiveStore.java
```diff
@@ -0,0 +1,189 @@
...
+
+  @Benchmark
+  @Threads(8)
+  public void selectiveStore_vector_128(ThreadLocalData tld) {
+    tld.conflict_vector_cnt = 0;
+    for (int i = 0; i < ARRAY_LENGTH; i += INT_128_SPECIES.length()) {
+      IntVector av = IntVector.fromArray(INT_128_SPECIES, tld.input1, i);
+      IntVector bv = IntVector.fromArray(INT_128_SPECIES, tld.input2, i);
+      IntVector cv = IntVector.fromArray(INT_128_SPECIES, tld.index, i);
+      VectorMask<Integer> mask = av.compare(VectorOperators.NE, bv);
```
jatin-bhateja, line 109: Can we mover vector compare out of benchmarking loop to reduce the noise.

JoshuaZhuwj: Thanks Jatin for your review.
This test I added is placed together with other benchmarks extracted from user cases.
The mask computation inside the loop is the essential logic in real sceanario.
We can add micro benchmarks in test/micro/org/openjdk/bench/jdk/incubator/vector/operation and then evaluate performance for compression / expanding in single API level.

jatin-bhateja: Ok got it, thanks! 
BTW can you also post the AVX2 numbers just to see how does it compare against scalar version when direct compress/expand instructions are not supported by target.

JoshuaZhuwj: FYI
image

theRealELiu:
> How about the performance for byte and short types? Do you have performance data comparing Java implementation against with scalar version? As some platforms may not support the compress/expand operation well. E.g., SVE doesn’t support compress on byte and short types. AVX512 maybe the same without vbmi2 extension.

I think the result of AVX2 could reflect the same issue. It's a shame that SVE misses the direct compress instruction for byte and short types.

JoshuaZhuwj:
> How about the performance for byte and short types? Do you have performance data comparing Java implementation against with scalar version? As some platforms may not support the compress/expand operation well. E.g., SVE doesn’t support compress on byte and short types. AVX512 maybe the same without vbmi2 extension.

Eric, on platforms without hardware support, like AVX2 or Neon, vector-to-vector compression could be a composition of mask -> shuffle -> rearrange. A precomputed permutation table could be leveraged to lookup shuffle from mask. John Rose elaborated the algorithm how to compute a compression shuffle from a mask. See design discussion at https://github.com/openjdk/panama-vector/pull/115

JoshuaZhuwj:
Thanks for all your review comments. I update the benchmark accordingly.

PR #157, opened 2021-10-28
8276083: Incremental patch to further optimize new compress/expand APIs over X86

jatin-bhateja:
Summary of changes:
1)  Added both scalar and vector variants of JMH performance tests for Vector.compress/expand and VectorMask.compress APIs.

2) Improved performance of operations where mask length is less than 4. Mask loading is a two stage process where in first the boolean array is loaded into a vector and then either transferred to a predicate register or a vector whose size is equivalent to that of underlined SPECIES.  A mask whose length is less than 4 will result into a less than 32 bit vector load operation.  Operations dependent on smaller masks are now being handled in java side implementation of these and some other APIs.  Since the condition for special handling and fallback logic leading to C2 intrinsic call is based on constant expression hence one of the control path is optimized out. This shall also prevent any performance penalty due to failed lazy inline expansion which most often occurs due to unsupported vector sizes.  If lazy inline expansion fails then C2 emits a direct call instruction to a callee method and thus we also loose any opportunity for procedure in-lining at that point, a separate [issue ](https://bugs.openjdk.java.net/browse/JDK-8276085)has been created to address this problem.
 
3) Improved performance of VectorMask.compress over legacy non-AVX512 targets, added the missing checks in C2Compiler::is_intrinsics_supported routine to enable procedure in-lining early during parsing if target does not support direct compress/expand instructions.  
    
4) Inline expand VectorMask.intoArray operation to trigger boxing-unboxing optimization. This significantly improved the performance of VectorMask.compress in newly added JMH micros.

Following is the performance data for included JMH micros:
System Configuration:  Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz (Icelake Server 40C 2S)

A) VectorMask.compress:

image

B) Vector.compress:

image

C) Vector.expand:

image

Patch has been regressed using tier3 regressions at various AVX levels 0/1/2/3/KNL.

Kindly review  and share your feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8276083](https://bugs.openjdk.java.net/browse/JDK-8276083): Incremental patch to further optimize new compress/expand APIs over X86

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/157/head:pull/157` \
`$ git checkout pull/157`

`$ git checkout pull/157` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/157/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 157`

`$ git pr show -t 157`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/157.diff">https://git.openjdk.java.net/panama-vector/pull/157.diff</a>

</details>

### DIFF
--- a/test/jdk/jdk/incubator/vector/templates/Perf-CompressExpand.template
+++ b/test/jdk/jdk/incubator/vector/templates/Perf-CompressExpand.template
@@ -39,7 +39,7 @@
         boolean[] rs = fmt.apply(size);
 
         for (int ic = 0; ic < INVOC_COUNT; ic++) {
-            for (int i = 0, j = 0; i < a.length; i += SPECIES.length()) {
+            for (int i = 0, j = 0; i < ms.length; i += SPECIES.length()) {
                 VectorMask<$Wideboxtype$> m = VectorMask.fromArray(SPECIES, ms, i);
                 m.compress().intoArray(rs, j);
                 j += m.trueCount();


jatin-bhateja:
@PaulSandoz , your comments are addressed. Please let me know if there are further comments.

PaulSandoz:
> @PaulSandoz , your comments are addressed. Please let me know if there are further comments.

Did you run `gen-tests.sh`? 

I tried running that on a clone of your branch and it generates more code, some of which does not compile:

1. `maskCompress` iterates over a non-existent array `a`.
2. Benchmarks for ROR and ROL are generated and the scalar benchmarks refer to non-existent methods `ROR_scalar` and `ROL_scalar`.
3. The gather/scatter benchmarks are deleted.

So it looks like we got out on sync. We should fix the latter two in `vectorIntrinsics` so you can merge that in.

jatin-bhateja:
> > @PaulSandoz , your comments are addressed. Please let me know if there are further comments.
> 
> Did you run `gen-tests.sh`?
> 
> I tried running that on a clone of your branch and it generates more code, some of which does not compile:
> 
> 1. `maskCompress` iterates over a non-existent array `a`.
> 2. Benchmarks for ROR and ROL are generated and the scalar benchmarks refer to non-existent methods `ROR_scalar` and `ROL_scalar`.
> 3. The gather/scatter benchmarks are deleted. (They got removed when i switched the scatter/gather unit tests over to the load/store files. See [8266518: Refactor and expand scatter/gather tests jdk17#48](https://github.com/openjdk/jdk17/pull/48))
> 
> So it looks like we got out of sync, likely via merges from mainline. We should at least fix 2 and possibly reconsider how to do 3.

Hi @PaulSandoz ,
I have fixed 1 it was a typo,  as already mentioned I did some manual editing since existing generation is broken.

I will post the fixes for items 2 and 3 in a immediate subsequent patch once this gets integrated, changes in this patch are related to compress/expand and its related performance improvements.

PaulSandoz:
Yes, we can fix 2/3 separately independently of this branch. In fact i think for 3 we should consider a separate template like with the unit tests.

PaulSandoz (approved):
Java changes are good.

jatin-bhateja:
Thanks @PaulSandoz. Integrating this patch.

PR #158, opened 2021-11-02
Improve mask reduction operations on AVX

merykitty:
Hi,
This patch improves the logic of vector mask reduction operations on AVX, especially int, float, long, double, by using vmovmskpd and vmovmskps instructions. I also do a little refactoring to reduce duplication in toLong. The patch temporarily disables these operations on Neon, though.
Thank you very much.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/158/head:pull/158` \
`$ git checkout pull/158`

`$ git checkout pull/158` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/158/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 158`

`$ git pr show -t 158`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/158.diff">https://git.openjdk.java.net/panama-vector/pull/158.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4054,39 +4054,38 @@ void C2_MacroAssembler::masked_op(int ideal_opc, int mask_len, KRegister dst,
 
 #ifdef _LP64
 void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, KRegister mask,
-                                              Register tmp, int masklen, int masksize,
-                                              int vec_enc) {
+                                              int masklen, int masksize, int vec_enc) {
+  assert(VM_Version::supports_bmi1(), "Strange hardware");
   if(VM_Version::supports_avx512bw()) {
-    kmovql(tmp, mask);
+    kmovql(dst, mask);
   } else {
     assert(masklen <= 16, "");
-    kmovwl(tmp, mask);
+    kmovwl(dst, mask);
   }
 
   // Mask generated out of partial vector comparisons/replicate/mask manipulation
   // operations needs to be clipped.
   if (masksize < 16 && opc != Op_VectorMaskFirstTrue) {
-    andq(tmp, (1 << masklen) - 1);
+    andq(dst, (1 << masklen) - 1);
   }
 
   switch(opc) {
     case Op_VectorMaskTrueCount:
-      popcntq(dst, tmp);
+      popcntq(dst, dst);
       break;
     case Op_VectorMaskLastTrue:
-      movl(dst, -1);
-      bsrq(tmp, tmp);
-      cmov32(Assembler::notZero, dst, tmp);
+      lzcntq(dst, dst);
+      notl(dst);
+      addl(dst, 64);
       break;
     case Op_VectorMaskFirstTrue:
-      assert(VM_Version::supports_bmi1(), "");
       if (masklen < 32) {
-        orl(tmp, 1 << masklen);
-        tzcntl(dst, tmp);
+        orl(dst, 1 << masklen);
+        tzcntl(dst, dst);
       } else if (masklen == 32) {
-        tzcntl(dst, tmp);
+        tzcntl(dst, dst);
       } else {
-        tzcntq(dst, tmp);
+        tzcntq(dst, dst);
       }
       break;
     case Op_VectorMaskToLong:
@@ -4146,9 +4145,15 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
       popcntl(dst, tmp);
       break;
     case Op_VectorMaskLastTrue:
-      movl(dst, -1);
-      bsrl(tmp, tmp);
-      cmov32(Assembler::notZero, dst, tmp);
+      if (VM_Version::supports_bmi1()) {
+        lzcntl(dst, tmp);
+        notl(dst);
+        addl(dst, 32);
+      } else {
+        movl(dst, -1);
+        bsrl(tmp, tmp);
+        cmov32(Assembler::notZero, dst, tmp);
+      }
       break;
     case Op_VectorMaskFirstTrue:
       if (VM_Version::supports_bmi1()) {
@@ -4170,6 +4175,7 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
       }
       break;
     case Op_VectorMaskToLong:
+      assert(dst == tmp, "Dst and tmp should be the same for toLong operations");
       break;
     default: assert(false, "Unhandled mask operation");
   }
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.hpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.hpp
@@ -224,7 +224,7 @@ public:
 
  public:
 #ifdef _LP64
-  void vector_mask_operation(int opc, Register dst, KRegister mask, Register tmp, int masklen, int masksize, int vec_enc);
+  void vector_mask_operation(int opc, Register dst, KRegister mask, int masklen, int masksize, int vec_enc);
 
   void vector_mask_operation(int opc, Register dst, XMMRegister mask, XMMRegister xtmp,
                              Register tmp, int masklen, BasicType bt, int vec_enc);
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -8652,7 +8652,7 @@ instruct vmask_tolong_evex(rRegL dst, kReg mask, rFlagsReg cr) %{
     int mask_len = Matcher::vector_length(this, $mask);
     int mask_size = mask_len * type2aelembytes(mbt);
     int vlen_enc = vector_length_encoding(this, $mask);
-    __ vector_mask_operation(opcode, noreg, $mask$$KRegister, $dst$$Register,
+    __ vector_mask_operation(opcode, $dst$$Register, $mask$$KRegister,
                              mask_len, mask_size, vlen_enc);
   %}
   ins_pipe( pipe_slow );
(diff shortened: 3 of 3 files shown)


PaulSandoz:
Thanks, this will require review from engineers more knowledgeable than I on the x86 aspects and ARM aspects. IMO we cannot integrate until there is a plan for ARM either as part of this PR or a follow on PR.

merykitty:
Microbenchmark shows significant improvement on my AVX2 machine.

    Before:
    Benchmark               Mode  Cnt   Score   Error  Units
    MaskReduction.byte128   avgt   25   0.773 ± 0.006  ns/op
    MaskReduction.byte256   avgt   25   0.778 ± 0.007  ns/op
    MaskReduction.int128    avgt   25   1.061 ± 0.008  ns/op
    MaskReduction.int256    avgt   25   1.553 ± 0.010  ns/op
    MaskReduction.long128   avgt   25  43.008 ± 0.354  ns/op
    MaskReduction.long256   avgt   25   1.271 ± 0.006  ns/op
    MaskReduction.short128  avgt   25   0.989 ± 0.006  ns/op
    MaskReduction.short256  avgt   25   0.919 ± 0.005  ns/op

    After:
    Benchmark               Mode  Cnt   Score   Error  Units
    MaskReduction.byte128   avgt   25   0.566 ± 0.001  ns/op
    MaskReduction.byte256   avgt   25   0.556 ± 0.003  ns/op
    MaskReduction.int128    avgt   25   0.553 ± 0.003  ns/op
    MaskReduction.int256    avgt   25   0.828 ± 0.002  ns/op
    MaskReduction.long128   avgt   25  41.618 ± 0.241  ns/op
    MaskReduction.long256   avgt   25   0.552 ± 0.001  ns/op
    MaskReduction.short128  avgt   25   0.775 ± 0.004  ns/op
    MaskReduction.short256  avgt   25   0.834 ± 0.006  ns/op

The benchmark simply load 2 constant vector, compare them and return the toLong of the result as follow:

    return BYTE_128_1.eq(BYTE_128_2).toLong();

merykitty:
Thank @PaulSandoz for the response, I hope this PR will be useful in the future.

Besides, I found out that C2 fails with `Long128Vector`, even a simple operation of `LONG_128_1.eq(LONG_128_2)`, is this a known behaviour?

PaulSandoz:
> Besides, I found out that C2 fails with `Long128Vector`, even a simple operation of `LONG_128_1.eq(LONG_128_2)`, is this a known behaviour?

I was not aware of that, @jatin-bhateja might know more.

merykitty:
It seems that `Long128Vector` disappoints C2 due to the fact that C2 doesn't know how to box a vector mask of length 2, which in turn is due to `StoreVectorNode` currently don't accept vectors of length 2. I will submit a fix for this later.

theRealELiu:
AArch64 needs some changes after this PR, I'm now investigating that.

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -698,9 +698,6 @@ bool LibraryCallKit::inline_vector_mask_operation() {
   ciKlass* mbox_klass = mask_klass->const_oop()->as_instance()->java_lang_Class_klass();
   const TypeInstPtr* mask_box_type = TypeInstPtr::make_exact(TypePtr::NotNull, mbox_klass);
   Node* mask_vec = unbox_vector(mask, mask_box_type, elem_bt, num_elem, true);
-  if (mask_vec->bottom_type()->isa_vectmask() == NULL) {
-    mask_vec = gvn().transform(VectorStoreMaskNode::make(gvn(), mask_vec, elem_bt, num_elem));
-  }
```
theRealELiu, line 703: There may have some regressions on AArch64 after refactoring those reduction operations base on this PR. 

With the `VectorStoreMaskNode`, the load-store pair could be optimized. After removing that, `VectorLoadMaskNode` needs extra extension instructions for various type. E.g.

```
After:
        ldr     h16, [x10, #16]  // LoadVector
        uxtl    v16.8h, v16.8b
        uxtl    v16.4s, v16.4h
        uxtl    v16.2d, v16.2s
        neg     v16.2d, v16.2d  // VectorLoadMask
        neg     v17.2d, v16.2d
        addv    b17, v17.16b
        umov    w11, v17.b[0]   // TrueCount
        sxtw    x0, w11         // StoreVector

Before:

        ldr     h16, [x10, #16] // LoadVector
        addv    b17, v16.8b
        umov    w11, v17.b[0]   // TrueCount
        sxtw    x0, w11         // StoreVector
```

No sliver bullet on this problem. Either you can match `VectorMaskTrueCount (VectorLoadMask src)` on AArch64 with this mid-end change in your patch, or just let the mid-end stay the same and match `VectorMaskTrueCount (VectorStoreMask src)` on x86 to pursue the better performance. The later would be better for its less effort.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4079,33 +4079,70 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, KRegister m
...
-      mov64(dst, -1);
-      bsrq(tmp, tmp);
-      cmov(Assembler::notZero, dst, tmp);
+      if (need_clip) {
+        // need_clip implies masklen < 32
+        andl(dst, (1 << masklen) - 1);
+      }
+      lzcntl(dst, dst);
+      negl(dst);
+      addl(dst, 31);
```
jatin-bhateja, line 4134: I think from latency perspective earlier sequence was   better given that  constant move to a register is not scheduled to an execution port.

jatin-bhateja (commented):
Over all changes looks ok to me.
- Savings are majorly because we are preventing additional vector store mask.
- Tier-3/4 testing at various AVX Levels 0/1/2/3/KNL went clean.

jatin-bhateja:
> > Besides, I found out that C2 fails with `Long128Vector`, even a simple operation of `LONG_128_1.eq(LONG_128_2)`, is this a known behaviour?
> 
> I was not aware of that, @jatin-bhateja might know more.

Yes, this the current [behavior ](https://github.com/openjdk/panama-vector/pull/157#discussion_r738549758)

src/hotspot/share/opto/vectorIntrinsics.cpp
merykitty: Thank you very much for your investigation, I have reverted the changes in the mid-end and pushed the change to the back-end as your suggestion. Another approach may be to elide the `VectorLoadMaskNode` attached to a mask reduction node but I think it should be another PR later.

merykitty:
The latest change reverts the change in the mid-end and lets the x86 back-end perform the elision when matching. Some minor changes in the C2_MacroAssembler were made to improve the code emitted by first true operations, as well as to reduce the width of instructions that operate on index calculations from 64 bits to 32 bits.
Thank you very much.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4062,50 +4062,114 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, KRegister m
...
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
       break;
     case Op_VectorMaskLastTrue:
-      mov64(dst, -1);
+      movl(dst, -1);
       bsrq(tmp, tmp);
-      cmov(Assembler::notZero, dst, tmp);
+      cmov32(Assembler::notZero, dst, tmp);
```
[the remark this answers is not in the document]
merykitty: Thank you very much for your reviews and suggestions. I have measured both approaches. While these provide similar performance on Intels with the `lzcnt` offers better throughput at the exchange of slightly worse latency. The situation is quite different on Amds due to the inefficiency of `bsf` in comparison with `lzcnt`. So I think using `lzcnt` where possible would be more preferable.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4146,9 +4145,15 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
       popcntl(dst, tmp);
       break;
     case Op_VectorMaskLastTrue:
-      movl(dst, -1);
-      bsrl(tmp, tmp);
-      cmov32(Assembler::notZero, dst, tmp);
+      if (VM_Version::supports_bmi1()) {
```
sviswa7, line 4148: This should be supports_lzcnt().

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4054,39 +4054,38 @@ void C2_MacroAssembler::masked_op(int ideal_opc, int mask_len, KRegister dst,
 
 #ifdef _LP64
 void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, KRegister mask,
-                                              Register tmp, int masklen, int masksize,
-                                              int vec_enc) {
+                                              int masklen, int masksize, int vec_enc) {
+  assert(VM_Version::supports_bmi1(), "Strange hardware");
```
sviswa7, line 4058: Also need to check for supports_lzcnt().

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
merykitty: Oh right, thanks for noticing.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
merykitty: We have separate checks for all `popcnt`, `lzcnt` and `tzcnt`, do you think the function should verify or just let the assembler throw after?

sviswa7: Good to have assert here in this function as well.

sviswa7:
@merykitty Please send this PR directly versus the openjdk/jdk mainline. This is a good optimization to have in JDK 18.

merykitty:
@sviswa7 I have addressed your reviews and created a pull request in mainline. May you create an issue for the PR, please.
Thank you very much.

sviswa7:
> @sviswa7 I have addressed your reviews and created a pull request in mainline. May you create an issue for the PR, please. Thank you very much.

The PR is:
https://bugs.openjdk.java.net/browse/JDK-8277426

PR #182, opened 2022-03-21
8283413: Add C2 mid-end and x86 back-end implementation for bit REVERSE and REVERSE_BYTES operation

jatin-bhateja:
Hi All,

Patch includes following changes:-
- New C2 IR nodes to support VectorOperations.REVERSE operation.
- X86 backend implementation for targets supporting AVX2, AVX512 and GFNI features.

Please find below the performance data of Vector API JMH micros:-

System Configuration:
ICX: Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz (40C 2S)
CLX: Intel(R) Xeon(R) Platinum 8280 CPU @ 2.70GHz (28C 2S)

image

Kindly review and share your feedback.

Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8283413](https://bugs.openjdk.java.net/browse/JDK-8283413): Add C2 mid-end and x86 back-end implementation for bit REVERSE and REVERSE_BYTES operation

### Reviewers
 * [Xiaohong Gong](https://openjdk.java.net/census#xgong) (@XiaohongGong - Committer) ⚠️ Review applies to 2c7815a470689d01c38859e2db9970acd3f5d584
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to a3a37a3c40430b613930decec2fe39603d30cf8d
 * [Eric Liu](https://openjdk.java.net/census#eliu) (@theRealELiu - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/182/head:pull/182` \
`$ git checkout pull/182`

`$ git checkout pull/182` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/182/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 182`

`$ git pr show -t 182`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/182.diff">https://git.openjdk.java.net/panama-vector/pull/182.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -3403,22 +3403,16 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
     break;
   }
 
+  case Op_ReverseBytesV:
   case Op_ReverseV: {
-    if (n->in(1)->Opcode() == Op_ReverseV) {
+    if ((uint)n->in(1)->Opcode() == nop) {
       if (n->is_predicated_vector() && n->in(1)->is_predicated_vector() &&
           n->in(2) == n->in(1)->in(2)) {
+        // Node (Node X , Mask) Mask => X
         n->subsume_by(n->in(1)->in(1), this);
-      } else {
-        bool is_user_blend = false;
-        for (DUIterator_Fast imax, i = n->fast_outs(imax); i < imax; i++) {
-          if (n->fast_out(i)->Opcode() == Op_VectorBlend) {
-            is_user_blend = true;
-            break;
-          }
-        }
-        if (!is_user_blend) {
-          n->subsume_by(n->in(1)->in(1), this);
-        }
+      } else if (!n->is_predicated_using_blend() && !n->in(1)->is_predicated_using_blend()) {
+        // Node (Node X) =>  X
+        n->subsume_by(n->in(1)->in(1), this);
       }
     }
     break;
--- a/src/hotspot/share/opto/node.hpp
+++ b/src/hotspot/share/opto/node.hpp
@@ -793,6 +793,7 @@ public:
     Flag_is_predicated_vector        = 1 << 15,
     Flag_for_post_loop_opts_igvn     = 1 << 16,
     Flag_is_removed_by_peephole      = 1 << 17,
+    Flag_is_predicated_using_blend   = 1 << 18,
     _last_flag                       = Flag_is_removed_by_peephole
   };
 
@@ -1009,6 +1010,8 @@ public:
   bool is_reduction() const { return (_flags & Flag_is_reduction) != 0; }
 
   bool is_predicated_vector() const { return (_flags & Flag_is_predicated_vector) != 0; }
+
+  bool is_predicated_using_blend() const { return (_flags & Flag_is_predicated_using_blend) != 0; }
 
   // The node is a CountedLoopEnd with a mask annotation so as to emit a restore context
   bool has_vector_mask_set() const { return (_flags & Flag_has_vector_mask_set) != 0; }
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -546,6 +546,7 @@ bool LibraryCallKit::inline_vector_nary_operation(int n) {
       operation->add_flag(Node::Flag_is_predicated_vector);
     } else {
       operation = gvn().transform(operation);
+      operation->add_flag(Node::Flag_is_predicated_using_blend);
       operation = new VectorBlendNode(opd1, operation, mask);
     }
   }


src/hotspot/share/opto/compile.cpp
```diff
@@ -3403,6 +3403,27 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
...
+            break;
+          }
+        }
+        if (!is_user_blend) {
+          n->subsume_by(n->in(1)->in(1), this);
+        }
+      }
+    }
+    break;
+  }
```
XiaohongGong, line 3425: The mid-end part looks good to me.  Thanks! BTW, may I ask why you add this transformation here while not in the `ReverseVNode::Idea()` ? Thanks so much！

jatin-bhateja: Blend IR node is created in a subsequent step after ReverseV node creation and are not visible to during eager identity transformation of ReverseV;   Final graph reshaping gives us opportunity to perform point transformation on graph patterns which cannot be easily handled through gvn based transformations.

XiaohongGong: I see. Thanks for your explanation!

XiaohongGong:
The mid-end part looks good to me! Thanks!

XiaohongGong (approved):

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4393,3 +4393,181 @@ void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src,
   kunpckdql(dst, tmp, tmp);
 }
 #endif
+
+void C2_MacroAssembler::vector_reverse_bit_avx(BasicType bt, XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
```
merykitty, line 4397: You can do a this bit reverse using lookup table on each nibble and oring the results, the pseudocode would look something like this

    lut = broadcasti128(0b0000, 0b1000, 0b0100, 0b1100, 0b0010, 0b1010, 0b0110, 0b1110, 0b0001, ...)
    mask = pbroadcastd(0x0f0f0f0f)
    hi = pand(src, mask)
    hi = pshufb(lut, hi)
    hi = pslld(hi, 4)
    lo = psrld(src, 4)
    lo = pand(lo, mask)
    lo = pshufb(lut, lo)
    dst = por(lo, hi)

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4393,3 +4393,181 @@ void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src,
...
+      vector_reverse_byte_evex(bt, dst, dst, xtmp, xnoreg, noreg, vec_enc);
+    } else {
+      assert(vec_enc < Assembler::AVX_512bit, "");
+      vector_reverse_byte_avx(bt, dst, dst, xtmp, vec_enc);
+    }
+  }
+}
+#endif
+
+void C2_MacroAssembler::vector_reverse_byte_avx(BasicType bt, XMMRegister dst, XMMRegister src,
```
merykitty, line 4498: Since this is an in-lane shuffle, can we just use `vpshufb` for this?
Thanks.

merykitty: We are doing an element-wise byte reverse not a whole vector byte reverse I think we don't need any cross-lane shuffle here.
Thanks.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: Yes, this can be tried out. I guess its inspired from LUT based revbits algorithm from Hacker's delight section 7-1.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: Correct!, I realized it after commenting, corrected my statement.

jatin-bhateja: But a 3 cycle sequence still looks better than loading from memory followed by shuffle unless we are saving on code size as depicted in VBMI sequence which offers a direct vpermb instruction.

jatin-bhateja: Removed incorrect comments. your comments seems valid here.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: @merykitty , your comments addressed.

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -9964,6 +9964,15 @@ void Assembler::evpternlogq(XMMRegister dst, int imm8, KRegister mask, XMMRegist
   emit_int8(imm8);
 }
 
+void Assembler::vgf2p8affineqb(XMMRegister dst, int imm8, XMMRegister src2, XMMRegister src3, int vector_len) {
+  assert(VM_Version::supports_gfni(), "requires GFNI support");
+  assert(VM_Version::supports_avx(), "");
+  InstructionAttr attributes(vector_len, /* vex_w */ true, /* legacy_mode */ false, /* no_mask_reg */ true, /* uses_vl */ true);
+  attributes.set_is_evex_instruction();
```
merykitty, line 9971: This is not an evex-exclusive instruction

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8973,6 +8978,77 @@ instruct vcompress_mask_reg_evex(kReg dst, kReg mask, rRegL rtmp1, rRegL rtmp2,
 
 #endif // _LP64
 
+// -------------------------------- Bit and Byte Reversal Vector Operations ------------------------
+instruct vreverse_reg_avx(vec dst, vec src, vec xtmp1, vec xtmp2, vec xtmp3, rRegI rtmp) %{
```
merykitty, line 8982: This should be `legVec` as this relies on the AVX encoding

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8973,6 +8978,77 @@ instruct vcompress_mask_reg_evex(kReg dst, kReg mask, rRegL rtmp1, rRegL rtmp2,
...
+    BasicType bt  = Matcher::vector_element_basic_type(this);
+    InternalAddress addr = $constantaddress(T_LONG, vreplicate_imm(T_LONG, 0x8040201008040201L, 1));
+    __ vector_reverse_bit_gfni(bt, $dst$$XMMRegister, $src$$XMMRegister, $xtmp$$XMMRegister,
+                               addr, $rtmp$$Register, vec_enc);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct vreverse_byte_reg(vec dst, vec src, rRegI rtmp) %{
+  predicate(VM_Version::supports_avx512bw() || Matcher::vector_length_in_bytes(n) < 64);
```
merykitty, line 9026: For `!VM_Version::supports_avx512bw` this needs `legVec`

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4529,3 +4529,143 @@ void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src,
   kunpckdql(dst, tmp, tmp);
 }
 #endif
+
+void C2_MacroAssembler::vector_reverse_bit(BasicType bt, XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
+                                           XMMRegister xtmp2, XMMRegister xtmp3, Register rtmp, int vec_enc) {
+  if (VM_Version::supports_avx512bw()) {
```
merykitty, line 4535: `VM_Version::supports_avx512bw() && (VM_Version::supports_avx512vl() || vec_enc == AVX_512bit)`

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8973,6 +8978,77 @@ instruct vcompress_mask_reg_evex(kReg dst, kReg mask, rRegL rtmp1, rRegL rtmp2,
...
+    int vec_enc = vector_length_encoding(this);
+    BasicType bt = Matcher::vector_element_basic_type(this);
+    __ vector_reverse_bit(bt, $dst$$XMMRegister, $src$$XMMRegister, $xtmp1$$XMMRegister,
+                          $xtmp2$$XMMRegister, $xtmp3$$XMMRegister, $rtmp$$Register, vec_enc);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct vreverse_reg_evex(vec dst, vec src, vec xtmp1, vec xtmp2, rRegI rtmp) %{
+  predicate((VM_Version::supports_avx512bw() || Matcher::vector_length_in_bytes(n) == 64) && !VM_Version::supports_gfni());
```
merykitty, line 8997: If vector length < 64 we need avx512vl here. I think this and the above should be `VM_Version::supports_avx512vlbw()` instead

merykitty (commented):
Looks good to me otherwise, thank you very much.

src/hotspot/cpu/x86/x86.ad
merykitty: I misremembered the vector operand declaration, please ignore this.

src/hotspot/cpu/x86/x86.ad
merykitty: And this, too. Thanks.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Even though there is no X86 target which supports BW but not VL,  but we still need to be granular in emitting correct predications since one can always configure VMs to enable selective target features.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: VPSHUFB used here needs AVX512VLBW feature in-order to operate over 128/256/512 bit vector. Thus for an EVEX encoded instruction limiting feature here is AVX512VLBW.

src/hotspot/share/opto/compile.cpp
```diff
@@ -3403,6 +3403,27 @@ void Compile::final_graph_reshaping_main_switch(Node* n, Final_Reshape_Counts& f
...
+      if (n->is_predicated_vector() && n->in(1)->is_predicated_vector() &&
+          n->in(2) == n->in(1)->in(2)) {
+        n->subsume_by(n->in(1)->in(1), this);
+      } else {
+        bool is_user_blend = false;
+        for (DUIterator_Fast imax, i = n->fast_outs(imax); i < imax; i++) {
+          if (n->fast_out(i)->Opcode() == Op_VectorBlend) {
+            is_user_blend = true;
+            break;
+          }
```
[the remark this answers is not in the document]
jatin-bhateja: Here we are trying to optimize following scenarios by looking point graph patterns. So if there is a match then transformation will go though else graph will not be modified.  
ReverseV (ReverseV X) =>  X
ReverseV (Reverse X , Mask) Mask => X

jatin-bhateja:
Hi @sviswa7 , your comments have been addressed.

sviswa7 (commented):
Rest of the patch looks good to me.

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -1686,4 +1686,19 @@ class CountTrailingZerosVNode : public VectorNode {
...
+  virtual int Opcode() const;
+};
+
+class ReverseBytesVNode : public VectorNode {
+public:
+  ReverseBytesVNode(Node* in, const TypeVect* vt)
+  : VectorNode(in, vt) {}
+
+  virtual int Opcode() const;
+};
```
theRealELiu, line 1703: Just a little tip: Can we add Identify for BYTE type, that is ReverseBytesV(x) ->x.

theRealELiu (approved):
LGTM.

sviswa7 (approved):

src/hotspot/share/opto/vectornode.hpp
jatin-bhateja: Thanks @theRealELiu ,  its handled on the lines of Revers transformation during final graph reshaping.

jatin-bhateja: FTR, Reverse[bytes] related transforms are special since they cannot be done locally in associated Identity routines since corresponding IR  nodes may be succeeded by blends which are generated after creation of Reverse[bytes] IR, thus GVN related routines for both blend and Reverse[bytes] IR may not be able to handle this transform,  strict pattern match based transform during final graph re-shaping looks promising here.

theRealELiu (approved):

src/hotspot/share/opto/compile.cpp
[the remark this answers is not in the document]
sviswa7: @jatin-bhateja Please ignore, noticed that this was fixed in your last commit.

theRealELiu:
@PaulSandoz Thanks for your feedback. I think it can be fixed with related AArch64 patch, it's coming soon.

src/hotspot/share/opto/vectornode.hpp
XiaohongGong: Hi @jatin-bhateja ，currently I'm looking at https://bugs.openjdk.org/browse/JDK-8290485, and noticed that you said the transformations for Reverse[bytes] are special that we cannot do the transformations when it is used by a VectorBlend. Could you please explain why? Per my understanding, the masked Reverse[bytes] with VectorBlend should also work, no matter  what changes are applied to the Reverse[bytes], since we only need to make sure the non-masked lanes  come from the first input, and the masked lanes from the second (Reverse[bytes]).  Please incorrect me if I misunderstood. Thanks a lot!

XiaohongGong: Do you mean the `Flag_is_predicated_using_blend` used in the Idea is used to prevent optimizations like `ReverseBytesV (ReverseBytesV X , MASK) , MASK =>  X` when `ReverseBytesV` is predicated but implemented with `VectorBlend` ? So if it is,  then I think the input of ReverseBytesV should be `VectorBlend X (ReverseBytesV X) MASK` while not `ReverseBytesV X`, right?  Thanks!

jatin-bhateja: Hi @XiaohongGong , May be I was not explicit, currently we are skipping any ideal transformation for operations on predicated operations implemented using blend and  flag is set prior to a GVN transform, thus it can be safely used in ReverseBytes ideal routine. Since earlier these was no such flag and Blend node was created after GVN transform so it was not possible to skip transform for ReverseBytes[2] node.

image

XiaohongGong: Yeah, I know that is used to skip the transformations. I'm just curious why we need to skip the optimizations if it is used by a `VectorBlend`. Did you met any issues or something others? Thanks! I want to add a transformation such as `(ReverseBytesV src) => src` if the basic element type is T_BYTE. I cannot see the reason that I need to also skip this transformation for VectorBlend used ReverseBytesV. Maybe we can continue the discussion after I created a PR for it. WDYT?

XiaohongGong: Please refer to the PR https://github.com/openjdk/jdk/pull/9565. Thanks a lot!

jatin-bhateja: > Yeah, I know that is used to skip the transformations. I'm just curious why we need to skip the optimizations if it is used by a `VectorBlend`. Did you met any issues or something others? Thanks! I want to add a transformation such as `(ReverseBytesV src) => src` if the basic element type is T_BYTE. I cannot see the reason that I need to also skip this transformation for VectorBlend used ReverseBytesV. Maybe we can continue the discussion after I created a PR for it. WDYT?

It was purely to skip matching over bigger graph pallet at that point, with blends pattern matching will be slightly complex and will need to be handled in VectorBlend idealization routines.

jatin-bhateja: I think we can have a separate PR to handle cascaded complementary masked operations over non-predicated targets to handle all the cases in one shot i.e. REVERSE, REVERSEBYTES, NOT.

Other interesting transforms could be  ROR X+ ROL X by same shift = X,   
complementary casts/re-interprets.,  X2Y + Y2X = X

PR #184, opened 2022-03-24
8283598: [vectorapi] Add new vector operation for compress and expand bits 

smita-kamath:
Hi,

I've added support for new vector operations for compressing bits of integral vector types(Byte/Short/Integer/Long).
The implementation is based on Compress or Generalized Extract mentioned in Hackers Delight by Henry S. Warren, Jr.
The implementation does the following: given a mask and the number to be compressed, the bits of the number corresponding to the set mask bit are selected and compressed. 

Currently, this PR addresses only Java changes for compress bits operation. I've also updated the test framework.
Do review and share feedback.

---------
### Progress

### Issue
 * [JDK-8283598](https://bugs.openjdk.java.net/browse/JDK-8283598): [vectorapi] Add new vector operation for compress and expand bits

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to e20c43e2d855edc365b82c5889890de379f0f1d6
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to 4ae7ffa3d9c1d418da366cf71e3f1a7821354022

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/184/head:pull/184` \
`$ git checkout pull/184`

`$ git checkout pull/184` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/184/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 184`

`$ git pr show -t 184`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/184.diff">https://git.openjdk.java.net/panama-vector/pull/184.diff</a>

</details>


src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
```diff
@@ -445,6 +445,32 @@ static byte rotateRight(byte a, int n) {
...
+        for (int i = 0; i < iters; i++) {
+            prefix_mask = (byte) (count_mask  ^ (count_mask  << 1)); // Parallel prefix
+            prefix_mask = (byte) (prefix_mask ^ (prefix_mask << 2));
+            prefix_mask = (byte) (prefix_mask ^ (prefix_mask << 4));
+            move_mask = (byte) (prefix_mask & bitmask); // Bits to move
+            bitmask = (byte)(bitmask ^ move_mask | (move_mask >> (1 << i))); // Compress bitmask
+            temp = (byte) (a & move_mask); // Bits of the number a to be moved.
+            a = (byte) (a ^ temp | (temp >> (1 << i))); // Compress a
+            count_mask = (byte) (count_mask & ~prefix_mask); // adjust count_mask by identifying bits that have 0 to the right
+        }
```
XiaohongGong, line 470: Hi @smita-kamath , can we simply count the bits number and then do a left shift on the result, e.g:
```
1) b =  a & bitmask
2) b = popcount(b)
3) result = (0x1 << b) - 1
```
Please correct me if I misunderstood something. Thanks so much!

merykitty:
Hi, what do you think about adding this operation for scalar first, and then adding it to vector api later. We could add several other bit manipulation operations such as `bitCount` and `numberOfLeadingZeros` for `Byte` and `Short` also.
Thanks.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
merykitty: @XiaohongGong  The compression is based on the bits of `bitmask`, not on the bits of `b`, so `bitcompress(0b1001, 0b1100) = 0b10`, `bitcompress(0b0101, 0b1110) = 0b010` (The number of bits in the result I write to be equal to the number of set bits in the mask for easier visualisation).

XiaohongGong: Oh, I see, thanks for the explanation!

smita-kamath:
Hi @merykitty, That is a good idea. However, the effort to bring it to scalar is much larger as we need to file a JEP and follow the required processes. This is why we plan to introduce it in Vector API first.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
```diff
@@ -445,6 +445,32 @@ static byte rotateRight(byte a, int n) {
         return (byte)(((((byte)a) & Byte.toUnsignedInt((byte)-1)) >>> (n & Byte.SIZE-1)) | ((((byte)a) & Byte.toUnsignedInt((byte)-1)) << (Byte.SIZE - (n & Byte.SIZE-1))));
     }
 
+    /* Implementation note: The implementation is based on Compress or Generalized Extract mentioned in
+     * Henry S. Warren, Jr's Hackers Delight, Addison Wesley, 2002.
+     */
+    static byte compressBits(byte a, byte bitmask) {
```
theRealELiu, line 451: According to the draft of 4th incubation(https://bugs.openjdk.java.net/browse/JDK-8280173), the `compress bit` should be similar to `vector mask compress`(CompressM). As per my understanding, the second argument `bitmask` should be an all true mask here. If so, it's also can be removed.

PaulSandoz: I think i mistook the intended implementation when writing the JEP.

Since there is no scalar equivalent the operation is currently under specified, and in this case I think the expand operation also makes sense so `compress(expand(x, m), m) = x`

PaulSandoz: Just talked with @sviswa7. I will update the JEP. Also, i will follow up on scalar integral implementations for compress/expand bits, where the `Byte`/`Short` implementations defer to that on `Integer`. Intrinsic scalar implementations can be implemented afterwards.

smita-kamath: Hi @PaulSandoz, do you think this implementation of compress bits is fine? I can plan to create a separate PR for expand bits operation. Do let me know. Thanks.

theRealELiu: > I think i mistook the intended implementation when writing the JEP.
> 
> Since there is no scalar equivalent the operation is currently under specified, and in this case I think the expand operation also makes sense so `compress(expand(x, m), m) = x`

If the true count of `m` is less than the significant bits of `x` , I think `x` will lost bits so that this equation doesn't work. E.g., m = 0b00000001, x = 0b11111111. It's 0b00000001 after expand(x, m) , and if compressed it back with the same bitmask, `m` will get 0b00000001.

src/hotspot/share/opto/mulnode.hpp
```diff
@@ -389,4 +389,12 @@ class MulAddS2INode : public Node {
   virtual uint ideal_reg() const { return Op_RegI; }
 };
 
+//------------------------------CompressBitsNode-------------------------------
+// CompressBits placeholder node
+class CompressBitsNode : public Node {
+public:
+  CompressBitsNode( Node *in1, Node *in2 ) : Node(0,in1,in2) {}
```
XiaohongGong, line 396: Style: the spaces are not needed after `"("` and before `")"`. And spaces are needed between the arguments. And `"*"` is better to be next close to the type. So suggest to modify as:
`CompressBitsNode(Node* in1, Node* in2) : Node(0, in1, in2) {}`

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
XiaohongGong: So I think it's better to comment an example to the scalar method or API definition.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -457,6 +457,53 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
...
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 8));
+#end[!byte]
+#if[intOrLong]
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 16));
+#if[long]
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 32));
+#end[long]
+#end[intOrLong]
+            move_mask = ($type$) (prefix_mask & bitmask); // Bits to move
+            bitmask = ($type$)(bitmask ^ move_mask | (move_mask >> (1 << i))); // Compress bitmask
```
XiaohongGong, line 500: style: space between `($type$)(bitmask ^ move_mask | ...)`

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -457,6 +457,53 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
...
+        // Move mask identifies the bits to be moved
+        // temp identifies the bits of the given number to be moved
+        $type$ prefix_mask, move_mask, temp;
+#if[intOrLong]
+#if[int]
+        int iters = 5;
+#else[int]
+        int iters = 6;
+#end[int]
+#end[intOrLong]
```
XiaohongGong, line 478: why not simply with:
```
#if[int]
        int iters = 5;
#end[int]
#if[long]
        int iters = 6;
#end[long]
```

test/jdk/jdk/incubator/vector/templates/Unit-header.template
```diff
@@ -1500,6 +1500,47 @@ public class $vectorteststype$ extends AbstractVectorTest {
...
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 8));
+#end[!byte]
+#if[intOrLong]
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 16));
+#if[long]
+            prefix_mask = ($type$) (prefix_mask ^ (prefix_mask << 32));
+#end[long]
+#end[intOrLong]
+            move_mask = ($type$) (prefix_mask & b);
+            b = ($type$)(b ^ move_mask | (move_mask >> (1 << i)));
```
XiaohongGong, line 1536: style: add one space between `($type$)(b ^ move_mask ...)`

test/jdk/jdk/incubator/vector/templates/Unit-header.template
```diff
@@ -1500,6 +1500,47 @@ public class $vectorteststype$ extends AbstractVectorTest {
 #end[intOrLong]
     }
 
+    static $type$ COMPRESSBITS_scalar($type$ a, $type$ b) {
+        $type$ prefix_mask, move_mask, temp;
+        a = ($type$) (a & b);
+        $type$ count_mask = ($type$) (~b << 1);
+        $type$ mp, mv, t;
```
XiaohongGong, line 1507: These are not used?

src/hotspot/share/prims/vectorSupport.cpp
```diff
@@ -529,6 +529,16 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
       }
       break;
     }
+    case VECTOR_OP_COMPRESS_BITS: {
+     switch (bt) {
+        case T_BYTE: // Returning Op_CompressBits for
+        case T_SHORT:// all types temporarily.
+        case T_INT:
+        case T_LONG: return Op_CompressBits;
+        default: fatal("COMPRESS_BITS: %s", type2name(bt));
```
XiaohongGong, line 538: The identity issue "three spaces before `"case"`". Seems the above bitcount and reverse have the same issue. So please fix them together. Thanks!

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
PaulSandoz: Yes, thanks, it only works for certain cases, i did not think it through carefully.

PaulSandoz:
Here's the issue for scalar compress and expand bits [JDK-8283892](https://bugs.openjdk.java.net/browse/JDK-8283892)

theRealELiu:
> Here's the issue for scalar compress and expand bits [JDK-8283892](https://bugs.openjdk.java.net/browse/JDK-8283892)

Thanks for the update. That makes more sense to me now.

src/hotspot/share/opto/mulnode.hpp
smita-kamath: Thank you. Will fix this.

test/jdk/jdk/incubator/vector/templates/Unit-header.template
smita-kamath: Thanks for pointing this out. Will fix it.

PaulSandoz (commented):
I recommend moving the scalar compress implementations out of the template and into say `VectorIntrinsics`, that will make it easier to replace later on.

I started work on the scalar implementations [here](https://github.com/openjdk/jdk/compare/master...PaulSandoz:compress-expand-bits?expand=1) based on what you did, but tried to make the code more idiomatic, so maybe copy those? (Note these have yet to be be tested.)

smita-kamath:
@PaulSandoz, sounds good.  Thank you.

sviswa7:
@smita-kamath Please update the summary at the top of the PR to reflect the latest:
   1. Both compress and expand are implemented
   2. Support for only int/long vectors
Also it will be good to update the title of JDK-8283598 to reflect that expand is also implemented as part of this PR.

jatin-bhateja:
> @smita-kamath Please update the summary at the top of the PR to reflect the latest:
> 
> 1. Both compress and expand are implemented
> 2. Support for only int/long vectors
>    Also it will be good to update the title of JDK-8283598 to reflect that expand is also implemented as part of this PR.

@sviswa7 , any specific reason to drop byte/short versions for these APIs

sviswa7:
@jatin-bhateja byte/short support for compress/expand were dropped from Vector API on advice from Paul Sandoz.
@smita-kamath Please bring in code from Paul's PR https://git.openjdk.java.net/jdk/pull/8115 and use Integer and Long compress expand.

test/jdk/jdk/incubator/vector/unit_tests.template
```diff
@@ -0,0 +1,8647 @@
+/*
```
PaulSandoz, line 1: This file should not be part of the commit, nor should `perf_scalar_tests.template` and `perf_tests.template`

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -457,6 +457,85 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
         return ($type$)((((($type$)a) & $Boxtype$.toUnsignedInt(($type$)-1)) >>> (n & $Boxtype$.SIZE-1)) | (((($type$)a) & $Boxtype$.toUnsignedInt(($type$)-1)) << ($Boxtype$.SIZE - (n & $Boxtype$.SIZE-1))));
 #end[intOrLong]
     }
+
+#if[intOrLong]
+    /* Implementation note: The implementation is based on Compress or Generalized Extract mentioned in
+     * Henry S. Warren, Jr's Hackers Delight, Addison Wesley, 2002.
+     */
+    static $type$ compressBits($type$ a, $type$ mask) {
```
PaulSandoz, line 465: No need to make these part of the template. Copy from https://github.com/openjdk/jdk/pull/8115 as Sandhya indicated and place them in a package private separate class, such as `CompressExpand`. Do the same for the tests and use directly. Then the code will be easier to replace.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/X-Vector.java.template
```diff
@@ -525,6 +525,7 @@ public abstract class $abstractvectortype$ extends AbstractVector<$Boxtype$> {
         return ($type$)((((($type$)a) & $Boxtype$.toUnsignedInt(($type$)-1)) >>> (n & $Boxtype$.SIZE-1)) | (((($type$)a) & $Boxtype$.toUnsignedInt(($type$)-1)) << ($Boxtype$.SIZE - (n & $Boxtype$.SIZE-1))));
 #end[intOrLong]
     }
+
```
sviswa7: Extra empty line could be removed from template.

test/jdk/jdk/incubator/vector/templates/Unit-header.template
```diff
@@ -1500,6 +1502,7 @@ public class $vectorteststype$ extends AbstractVectorTest {
 #end[intOrLong]
     }
 
+
```
sviswa7: Extra empty line could be removed from template.

sviswa7 (approved):
Rest of the patch looks good to me.

test/jdk/jdk/incubator/vector/gen-template.sh
```diff
@@ -403,6 +403,8 @@ gen_binary_alu_op "AND_NOT" "a \& ~b" "BITWISE"
 gen_binary_alu_op "OR+or"    "a | b"   "BITWISE"
 # Missing:        "OR_UNCHECKED"
 gen_binary_alu_op "XOR"   "a ^ b"   "BITWISE"
+gen_binary_alu_op "COMPRESS_BITS" "CompressExpandTest.compress(a,b)" "intOrLong"
+gen_binary_alu_op "EXPAND_BITS" "CompressExpandTest.expand(a,b)" "intOrLong"
```
PaulSandoz, line 407:
```suggestion
gen_binary_alu_op "COMPRESS_BITS" "CompressExpandTest.compress(a, b)" "intOrLong"
gen_binary_alu_op "EXPAND_BITS" "CompressExpandTest.expand(a, b)" "intOrLong"
```

PaulSandoz (approved):

PR #185, opened 2022-03-25
8283709: Add x86 back-end implementation for bit BIT_COUNT operation

jatin-bhateja:
Summary of changes:

- Patch re-uses existing C2 IR nodes and re-organizes LUT based JIT code sequence of VectorOperations.BIT_COUNT operation 
  for sub-word type (BYTE, SHORT) vectors over X86 targets supporting AVA2 and AVX512 features.
- Efficient single instruction POPCOUNT instruction is emitted for applicable targets. 

Kindly review and share you feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8283709](https://bugs.openjdk.java.net/browse/JDK-8283709): Add x86 back-end implementation for bit BIT_COUNT operation

### Reviewers
 * [Xiaohong Gong](https://openjdk.java.net/census#xgong) (@XiaohongGong - Committer) ⚠️ Review applies to fce9fcbfde64c7fa109edeff9b9718516f7faaaf
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - Committer) ⚠️ Review applies to b4ea07937755731f7080c821e2a77b0db3906645

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/185/head:pull/185` \
`$ git checkout pull/185`

`$ git checkout pull/185` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/185/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 185`

`$ git pr show -t 185`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/185.diff">https://git.openjdk.java.net/panama-vector/pull/185.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4538,7 +4538,6 @@ void C2_MacroAssembler::vector_popcount_integral(BasicType bt, XMMRegister dst, 
 void C2_MacroAssembler::vector_popcount_integral_evex(BasicType bt, XMMRegister dst, XMMRegister src,
                                                       KRegister mask, bool merge, int vec_enc) {
   assert(VM_Version::supports_avx512vl() || vec_enc == Assembler::AVX_512bit, "");
-  assert(UsePopCountInstruction, "");
   switch(bt) {
     case T_LONG:
       assert(VM_Version::supports_avx512_vpopcntdq(), "");
--- a/src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
@@ -4124,7 +4124,7 @@ class StubGenerator: public StubCodeGenerator {
     StubRoutines::x86::_vector_reverse_byte_perm_mask_int = generate_vector_reverse_byte_perm_mask_int("perm_mask_int");
     StubRoutines::x86::_vector_reverse_byte_perm_mask_short = generate_vector_reverse_byte_perm_mask_short("perm_mask_short");
 
-    if (VM_Version::supports_avx2() && (!VM_Version::supports_avx512_vpopcntdq() || !UsePopCountInstruction)) {
+    if (VM_Version::supports_avx2() && !VM_Version::supports_avx512_vpopcntdq()) {
       // lut implementation influenced by counting 1s algorithm from section 5-1 of Hackers' Delight.
       StubRoutines::x86::_vector_popcount_lut = generate_popcount_avx_lut("popcount_lut");
     }
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -7792,7 +7792,7 @@ address generate_avx_ghash_processBlocks() {
     StubRoutines::x86::_vector_reverse_byte_perm_mask_int = generate_vector_reverse_byte_perm_mask_int("perm_mask_int");
     StubRoutines::x86::_vector_reverse_byte_perm_mask_short = generate_vector_reverse_byte_perm_mask_short("perm_mask_short");
 
-    if (VM_Version::supports_avx2() && (!VM_Version::supports_avx512_vpopcntdq() || !UsePopCountInstruction)) {
+    if (VM_Version::supports_avx2() && !VM_Version::supports_avx512_vpopcntdq()) {
       // lut implementation influenced by counting 1s algorithm from section 5-1 of Hackers' Delight.
       StubRoutines::x86::_vector_popcount_lut = generate_popcount_avx_lut("popcount_lut");
     }
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -1702,7 +1702,7 @@ void VM_Version::get_processor_features() {
   }
 
   // Use population count instruction if available.
-  if (supports_popcnt() || supports_avx512_vpopcntdq() || supports_avx512_bitalg()) {
+  if (supports_popcnt()) {
     if (FLAG_IS_DEFAULT(UsePopCountInstruction)) {
       UsePopCountInstruction = true;
     }
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -8628,11 +8628,10 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
 
 // --------------------------------- PopCount --------------------------------------
 instruct vpopcount_integral_reg_evex(vec dst, vec src) %{
-  predicate(UsePopCountInstruction &&
-            ((is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
+  predicate((is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
               VM_Version::supports_avx512_bitalg()) ||
              (is_non_subword_integral_type(Matcher::vector_element_basic_type(n->in(1))) &&
-              VM_Version::supports_avx512_vpopcntdq())));
+              VM_Version::supports_avx512_vpopcntdq()));
   match(Set dst (PopCountVI src));
   match(Set dst (PopCountVL src));
   ins_cost(400);
@@ -8653,11 +8652,10 @@ instruct vpopcount_integral_reg_evex(vec dst, vec src) %{
 %}
 
 instruct vpopcount_integral_reg_evex_masked(vec dst, vec src, kReg mask) %{
-  predicate(UsePopCountInstruction &&
-            ((is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
+  predicate((is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
               VM_Version::supports_avx512_bitalg()) ||
              (is_non_subword_integral_type(Matcher::vector_element_basic_type(n->in(1))) &&
-              VM_Version::supports_avx512_vpopcntdq())));
+              VM_Version::supports_avx512_vpopcntdq()));
   match(Set dst (PopCountVI src mask));
   match(Set dst (PopCountVL src mask));
   format %{ "vector_popcount_integral_evex_masked $dst, $src, $mask" %}
@@ -8672,8 +8670,7 @@ instruct vpopcount_integral_reg_evex_masked(vec dst, vec src, kReg mask) %{
 %}
 
 instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) %{
-  predicate(!UsePopCountInstruction ||
-            (!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
+  predicate((!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
             (!VM_Version::supports_avx512_bitalg() && is_subword_type(Matcher::vector_element_basic_type(n->in(1)))));
   match(Set dst (PopCountVI src));
   effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP rtmp);
@@ -8688,7 +8685,7 @@ instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) 
 %}
 
 instruct vpopcountL_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, vec xtmp3, rRegP rtmp) %{
-  predicate(!UsePopCountInstruction || !VM_Version::supports_avx512_vpopcntdq());
+  predicate(!VM_Version::supports_avx512_vpopcntdq());
   match(Set dst (PopCountVL src));
   effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP xtmp3, TEMP rtmp);
   format %{ "vector_popcount_long  $dst, $src\t! using $xtmp1, $xtmp2, $xtmp3, and $rtmp as TEMP" %}


src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4464,64 +4464,99 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
...
+    vpxorq(xtmp2, xtmp2, xtmp2, vec_enc);
+    vpsadbw(dst, xtmp1, xtmp2, vec_enc);
   } else {
     // We do not see any performance benefit of running
     // above instruction sequence on 256 bit vector which
     // can operate over maximum 4 long elements.
     ShouldNotReachHere();
   }
-  if (bt == T_INT) {
+  if (rbt == T_INT) {
```
XiaohongGong, line 4537: So the SLP and the VectorAPI uses the same `PopCountVL` with different basic type?

XiaohongGong: Is it possible to separate this cast for SLP and always uses "T_LONG' as the basic type for `PopCountVL` ?

jatin-bhateja: We can have type homogeneity for PopCountVL across VectorAPI and SLP once we start auto-vectorizing ConvL2I  operation.

nsjian: I think that makes the code hard to maintain, and I think @merykitty 's unaddressed comment in: https://github.com/openjdk/jdk/pull/6857#discussion_r780917717 is a preferable approach. 

FYI: @fg1417 has a ConvL2I auto-vectorization support under review: https://github.com/openjdk/jdk/pull/7806.

XiaohongGong: Yeah, so if the `ConvL2I` is supported to be auto-vectorized after then, the `PopCountL` is better to be auto-vectorized to "PopCountVL + VectorCastL2X" ? Do you have any plan on it? Thanks so much！

jatin-bhateja: > Yeah, so if the `ConvL2I` is supported to be auto-vectorized after then, the `PopCountL` is better to be auto-vectorized to "PopCountVL + VectorCastL2X" ? Do you have any plan on it? Thanks so much！

Agree, with your proposal.

XiaohongGong: OK, thanks!

XiaohongGong:
The mid-end part looks good to me! Thanks!

XiaohongGong (approved):

jatin-bhateja:
Patch needs to be updated with predicated patterns.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: > > Yeah, so if the `ConvL2I` is supported to be auto-vectorized after then, the `PopCountL` is better to be auto-vectorized to "PopCountVL + VectorCastL2X" ? Do you have any plan on it? Thanks so much！
> 
> Agree to your proposal, once we have auto-vectorization for ConvL2I n place we can change the IR.

Here is an interim patch which type casts the result of PopCountVL into integer vector and thus prevent any special 
handling in target specific backend implementations. But it seems more appropriate to generate correct IR snippet (PopCountL + ConvL2I) during scalar operation inline expansion and explicitly handle ConvL2I during auto-vectorization.  

```
diff --git a/src/hotspot/share/opto/superword.cpp b/src/hotspot/share/opto/superword.cpp
index 1bf4c7a3282..b3dd90d1867 100644
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -2549,11 +2549,21 @@ void SuperWord::output() {
           vn = VectorNode::make(opc, in1, in2, vlen, velt_basic_type(n));
           vlen_in_bytes = vn->as_Vector()->length_in_bytes();
         }
+      } else if (opc == Op_PopCountL) {
+        // TODO: Succeed PopCountL by ConvL2I during initial graph construction 
+        // once auto-vectorizer supports ConvL2I operation.
+        assert(n->req() == 2, "only one input expected");
+        Node* in = vector_opd(p, 1);
+  	   BasicType bt = in->bottom_type()->is_vect()->element_basic_type();
+         in = VectorNode::make(opc, in, NULL, vlen, bt);
+         _igvn.register_new_node_with_optimizer(in);
+	   vn = VectorCastNode::make(Op_VectorCastL2X, in, T_INT, vlen);
+         vlen_in_bytes = vn->as_Vector()->length_in_bytes();
       } else if (opc == Op_SqrtF || opc == Op_SqrtD ||
                  opc == Op_AbsF || opc == Op_AbsD ||
                  opc == Op_AbsI || opc == Op_AbsL ||
                  opc == Op_NegF || opc == Op_NegD ||
-                 opc == Op_PopCountI || opc == Op_PopCountL) {
+                opc == Op_PopCountI) {
         assert(n->req() == 2, "only one input expected");
         Node* in = vector_opd(p, 1);
         vn = VectorNode::make(opc, in, NULL, vlen, velt_basic_type(n));
```

XiaohongGong: Yeah, thanks for it. I will do it once PR https://github.com/openjdk/jdk/pull/7806 is merged. Thanks!

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4433,6 +4433,16 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
   }
 }
 
+void C2_MacroAssembler::vbroadcastd(XMMRegister dst, Register rtmp, int imm32, int vec_enc) {
```
sviswa7, line 4436: Good to have rtmp after imm32. Basically temp registers after required inputs.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4463,62 +4473,98 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
...
-    evmovdqul(xtmp2, k0, ExternalAddress(StubRoutines::x86::vector_popcount_lut()), true, vec_enc, rtmp);
-    vpandq(xtmp3, src, xtmp1, vec_enc);
-    vpshufb(xtmp3, xtmp2, xtmp3, vec_enc);
-    vpsrlw(dst, src, 4, vec_enc);
-    vpandq(dst, dst, xtmp1, vec_enc);
-    vpshufb(dst, xtmp2, dst, vec_enc);
-    vpaddb(xtmp3, dst, xtmp3, vec_enc);
-    vpxorq(xtmp1, xtmp1, xtmp1, vec_enc);
-    vpsadbw(dst, xtmp3, xtmp1, vec_enc);
+                                            XMMRegister xtmp2, Register rtmp, int vec_enc) {
```
sviswa7, line 4478: Comments need updating. Majority of comments above this method need to move to vector_popcount_byte. And then it is better to give comments at each step below in the method for easy review and maintenance. Please add comments to short, byte, long as well on similar lines.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4463,62 +4473,98 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
...
+                                             XMMRegister xtmp2, Register rtmp, int vec_enc) {
+  if (vec_enc == Assembler::AVX_512bit) {
+    vector_popcount_byte(xtmp1, src, dst, xtmp2, rtmp, vec_enc);
+    vpxorq(xtmp2, xtmp2, xtmp2, vec_enc);
+    vpsadbw(dst, xtmp1, xtmp2, vec_enc);
   } else {
     // We do not see any performance benefit of running
     // above instruction sequence on 256 bit vector which
     // can operate over maximum 4 long elements.
     ShouldNotReachHere();
```
sviswa7, line 4520: From Vector API point of view it would be good to have the intrinsic for 256-bit/128-bit Long vectors as well.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -2049,6 +2050,19 @@ const bool Matcher::match_rule_supported_vector_masked(int opcode, int vlen, Bas
...
+      if ((is_subword_type(bt) && !VM_Version::supports_avx512_bitalg()) ||
+          ((bt == T_INT) && !VM_Version::supports_avx512_vpopcntdq())) {
+        return false;
+      }
+      return true;
+
+    case Op_PopCountVL:
+      if ((bt == T_LONG) && !VM_Version::supports_avx512_vpopcntdq()) {
+        return false;
+      }
```
sviswa7, line 2063: bt is T_LONG for this case, we don't need to check type.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
 // --------------------------------- PopCount --------------------------------------
 
-instruct vpopcountI_popcntd(vec dst, vec src) %{
-  predicate(VM_Version::supports_avx512_vpopcntdq());
+instruct vpopcount_integral_reg_evex(vec dst, vec src) %{
+  predicate((((Matcher::vector_element_basic_type(n->in(1)) == T_INT ||
+               Matcher::vector_element_basic_type(n->in(1)) == T_LONG)) &&
+             VM_Version::supports_avx512_vpopcntdq()) ||
+            (is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
+             VM_Version::supports_avx512_bitalg()));
```
sviswa7, line 8637: This sort of check is happening multiple times. Should we have a function which takes bt and tells us if vpopcnt instruction is supported for the type.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
   match(Set dst (PopCountVI src));
-  format %{ "vector_popcount_int $dst, $src\t! vector popcount packedI" %}
+  match(Set dst (PopCountVL src));
+  ins_cost(400);
+  format %{ "vector_popcount_integral_evex $dst, $src" %}
   ins_encode %{
     assert(UsePopCountInstruction, "not enabled");
```
sviswa7, line 8643: UsePopCountInstruction is ambiguous now? subword vs int/long depend on different platform feature so cannot be combined?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
-    int vlen_enc = vector_length_encoding(this);
-    __ vector_popcount_int($dst$$XMMRegister, $src$$XMMRegister, xnoreg, xnoreg, xnoreg, noreg, vlen_enc);
+    int opcode = this->ideal_Opcode();
+    int vlen_enc = vector_length_encoding(this, $src);
+    BasicType bt = Matcher::vector_element_basic_type(this, $src);
+    __ vector_popcount_integral_evex(bt, $dst$$XMMRegister, $src$$XMMRegister, k0, true, vlen_enc);
+    // TODO: Once auto-vectorizer supports ConvL2I operation, PopCountVL
+    // should be succeeded by its corresponding vector IR and following
+    // special handling should be removed.
+    if (opcode == Op_PopCountVL && Matcher::vector_element_basic_type(this) == T_INT) {
```
sviswa7, line 8651: This needs clarification in comments that the behavior is different based on vector api vs auto vectorizer. Have we tested this to work appropriately in both cases?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
   ins_encode %{
     assert(UsePopCountInstruction, "not enabled");
-    int vlen_enc = vector_length_encoding(this);
-    __ vector_popcount_int($dst$$XMMRegister, $src$$XMMRegister, $xtmp1$$XMMRegister, $xtmp2$$XMMRegister,
-                           $xtmp3$$XMMRegister, $rtmp$$Register, vlen_enc);
+    int opcode = this->ideal_Opcode();
+    int vlen_enc = vector_length_encoding(this, $src);
+    BasicType bt = Matcher::vector_element_basic_type(this, $src);
+    __ evmovdquq($dst$$XMMRegister, $src$$XMMRegister, vlen_enc);
+    __ vector_popcount_integral_evex(bt, $dst$$XMMRegister, $src$$XMMRegister, $mask$$KRegister, true, vlen_enc);
```
sviswa7, line 8673: No auto vectorizer path here where the result in int vector for long vector.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
-  match(Set dst (PopCountVL src));
-  format %{ "vector_popcount_long  $dst, $src\t! vector popcount packedL" %}
+instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) %{
+  predicate((!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
+            (!VM_Version::supports_avx512_bitalg() && is_subword_type(Matcher::vector_element_basic_type(n->in(1)))));
+  match(Set dst (PopCountVI src));
+  effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP rtmp);
+  format %{ "vector_popcount_int  $dst, $src\t! using $xtmp1, $xtmp2 and $rtmp as TEMP" %}
   ins_encode %{
     assert(UsePopCountInstruction, "not enabled");
```
sviswa7, line 8685: The assert needs to be removed on this path.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
 
-instruct vpopcountL(vec dst, vec src, vec xtmp1, vec xtmp2, vec xtmp3, rRegP rtmp, rFlagsReg cc) %{
+instruct vpopcountL_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, vec xtmp3, rRegP rtmp) %{
   predicate(!VM_Version::supports_avx512_vpopcntdq());
   match(Set dst (PopCountVL src));
-  effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP xtmp3, TEMP rtmp, KILL cc);
+  effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP xtmp3, TEMP rtmp);
   format %{ "vector_popcount_long  $dst, $src\t! using $xtmp1, $xtmp2, $xtmp3, and $rtmp as TEMP" %}
   ins_encode %{
     assert(UsePopCountInstruction, "not enabled");
```
sviswa7, line 8700: The assert needs to be removed on this path. Do you know why the testing didn't catch this?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8615,56 +8629,85 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
     BasicType bt = Matcher::vector_element_basic_type(this);
-    __ vector_popcount_long(bt, $dst$$XMMRegister, $src$$XMMRegister, $xtmp1$$XMMRegister, $xtmp2$$XMMRegister,
-                            $xtmp3$$XMMRegister, $rtmp$$Register, vlen_enc);
+    __ vector_popcount_long($dst$$XMMRegister, $src$$XMMRegister, $xtmp1$$XMMRegister,
+                            $xtmp2$$XMMRegister, $rtmp$$Register, vlen_enc);
+    // TODO: Once auto-vectorizer supports ConvL2I operation, PopCountVL
+    // should be succeeded by its corresponding vector IR and following
+    // special handling should be removed.
+    if (bt == T_INT) {
+      __ evpmovqd($dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
```
sviswa7, line 8709: why are we using evex instruction for avx path?

jatin-bhateja: AVX512 targets not supporting AVX512_POPCOUNTD feature can still have 64 byte vector operation, since long pop-count was only enabled for 512 bit vectors uptill now hence a down-casting EVEX instruction worked fine.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Same a above, it should be OK to keep this as is since its already used in match_rule_supported method.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Same a above, it should be OK to keep this as is since its already used in match_rule_supported method.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Yes, because this is predicated operation pattern and auto-vectorizer does not support predicate operation inferencing.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Yes, we do have explicit bit count tests in for auto-vectorizer flow to test this flow.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: It a JVM flag, fall back code is still  based on Integer.bitCount and Long.bitCount APIs,  if the flag is turned off we do not intrinsify these scalar APIs.  match_rule_supported which is called through vector inline expander is also sensitive to this flag.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: This is only in couple of predicates, so it think it will not hurt much.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: New code has been modularized byte vector popcount is the leaf level operation and all the other primitive type operation build on top of it.

src/hotspot/cpu/x86/x86.ad
sviswa7: The post loop work on the mainline would support the masked instructions generation.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
sviswa7: Without comments around the instructions it is very hard to review what the set of instructions are supposed to do.

src/hotspot/cpu/x86/x86.ad
sviswa7: It is about maintenance and review.

src/hotspot/cpu/x86/x86.ad
sviswa7: But this is !VM_Version::supports_avx512_vpopcntdq() so UsePopCountInstruction should have been false.

src/hotspot/cpu/x86/x86.ad
sviswa7: I see you fixed this is in your review comment resolution by adding a check for avx512vl  around this instruction and a separate instruction generation for non avx512 path, thanks.

src/hotspot/cpu/x86/x86.ad
sviswa7: match_rule_supported is explicitly checking for avx512popcntdq and avx512bitalg feature. UsePopCountInstruction is only set based on avx512popcntdq feature.

sviswa7:
Other than couple of comments above, rest of the patch looks good to me.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Auto vectorizer restricts predication to load and store operations.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1702,7 +1702,7 @@ void VM_Version::get_processor_features() {
   }
 
   // Use population count instruction if available.
-  if (supports_popcnt()) {
+  if (supports_popcnt() || supports_avx512_vpopcntdq() || supports_avx512_bitalg()) {
```
sviswa7, line 1705: The UsePopCountInstruction is only for scalar popcont. Extending it to Vector popcount is causing lot of confustion. Let us keep it for scalar only as below:
if (supports_popcnt()) {
 ...
}

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8661,16 +8653,15 @@ instruct vpopcount_integral_reg_evex(vec dst, vec src) %{
...
-  predicate((((Matcher::vector_element_basic_type(n->in(1)) == T_INT ||
-               Matcher::vector_element_basic_type(n->in(1)) == T_LONG)) &&
-             VM_Version::supports_avx512_vpopcntdq()) ||
-            (is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
-             VM_Version::supports_avx512_bitalg()));
+  predicate(UsePopCountInstruction &&
+            ((is_subword_type(Matcher::vector_element_basic_type(n->in(1))) &&
+              VM_Version::supports_avx512_bitalg()) ||
+             (is_non_subword_integral_type(Matcher::vector_element_basic_type(n->in(1))) &&
+              VM_Version::supports_avx512_vpopcntdq())));
```
sviswa7, line 8660: Could be replaced by:
predicate(is_pop_count_instr_target(Matcher::vector_element_basic_type(n->in(1))));
Also no need to check for UsePopCountInstruction any where as it is only meant for scalar popcount.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8633,19 +8627,17 @@ instruct vmuladdaddS2I_reg(vec dst, vec src1, vec src2) %{
...
```
sviswa7, line 8635: Please replace by:
predicate(is_pop_count_instr_target(Matcher::vector_element_basic_type(n->in(1))));
Also no need to check for UsePopCountInstruction any where as it is only meant for scalar popcount.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8681,13 +8672,13 @@ instruct vpopcount_integral_reg_evex_masked(vec dst, vec src, kReg mask) %{
 %}
 
 instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) %{
-  predicate((!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
+  predicate(!UsePopCountInstruction ||
+            (!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
             (!VM_Version::supports_avx512_bitalg() && is_subword_type(Matcher::vector_element_basic_type(n->in(1)))));
```
sviswa7, line 8677: Could be replaced by:
predicate((Matcher::vector_element_basic_type(n->in(1)) != T_LONG) 
 && !is_pop_count_instr_target(Matcher::vector_element_basic_type(n->in(1))));

src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
```diff
@@ -4124,7 +4124,7 @@ class StubGenerator: public StubCodeGenerator {
     StubRoutines::x86::_vector_reverse_byte_perm_mask_int = generate_vector_reverse_byte_perm_mask_int("perm_mask_int");
     StubRoutines::x86::_vector_reverse_byte_perm_mask_short = generate_vector_reverse_byte_perm_mask_short("perm_mask_short");
 
-    if (UsePopCountInstruction && VM_Version::supports_avx2() && !VM_Version::supports_avx512_vpopcntdq()) {
+    if (VM_Version::supports_avx2() && (!VM_Version::supports_avx512_vpopcntdq() || !UsePopCountInstruction)) {
```
sviswa7, line 4127: Please remove the reference to UsePopCountInstruction here.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -7792,7 +7792,7 @@ address generate_avx_ghash_processBlocks() {
     StubRoutines::x86::_vector_reverse_byte_perm_mask_int = generate_vector_reverse_byte_perm_mask_int("perm_mask_int");
     StubRoutines::x86::_vector_reverse_byte_perm_mask_short = generate_vector_reverse_byte_perm_mask_short("perm_mask_short");
 
-    if (UsePopCountInstruction && VM_Version::supports_avx2() && !VM_Version::supports_avx512_vpopcntdq()) {
+    if (VM_Version::supports_avx2() && (!VM_Version::supports_avx512_vpopcntdq() || !UsePopCountInstruction)) {
```
sviswa7, line 7795: Please remove the reference to UsePopCountInstruction here.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8697,12 +8688,11 @@ instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp)
 %}
 
 instruct vpopcountL_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, vec xtmp3, rRegP rtmp) %{
-  predicate(!VM_Version::supports_avx512_vpopcntdq());
+  predicate(!UsePopCountInstruction || !VM_Version::supports_avx512_vpopcntdq());
```
sviswa7, line 8691: Please remove the reference to UsePopCountInstruction here.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4535,10 +4535,10 @@ void C2_MacroAssembler::vector_popcount_integral(BasicType bt, XMMRegister dst,
   }
 }
 
-
 void C2_MacroAssembler::vector_popcount_integral_evex(BasicType bt, XMMRegister dst, XMMRegister src,
                                                       KRegister mask, bool merge, int vec_enc) {
   assert(VM_Version::supports_avx512vl() || vec_enc == Assembler::AVX_512bit, "");
+  assert(UsePopCountInstruction, "");
```
sviswa7, line 4541: Please remove this assert on UsePopCountInstruction.

sviswa7 (approved):
Please make the changes suggested here. No need for re-review.

src/hotspot/cpu/x86/vm_version_x86.cpp
jatin-bhateja: UsePopcountInstruction implies generating direct popcount instruction for targets which support it. A bit count operation can also be implemented without popcount as depicted by this patch.

jatin-bhateja: I agree its adding some confusion since user may turn off UsePopcountInstruction and may want to generate optimized instructions for bit count over AVX2 targets using Vector API.

For SLP  turning off UsePopcountInstruction will prevent creating scalar IR which will also inhabit interacting of vector operation by auto vectorizer.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: ADLC does not include code written in %source { }  clause .ad file in to generate DFA.cpp
And adding a prediction logic into a new matcher routine or some other files included by dfa.cpp looks like an overkill since its appearing on only two instruction patterns.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: There is still a matcher path which can be chosen even if user turns off UsePopCountInstruction,  relating this flag to an operation may not be correct.  But I agree that to avoid any confusion we can remove its usage from pattern for the time being.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: Again, its counter intuitive to enter into this control flow with UsePopcountInstruction being false.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Thanks,  I already cleaned this up, its used at couple of places and will refactor when I create mainline integration patch.

PR #187, opened 2022-03-29
Load store memory segment

PaulSandoz:
This is the initial implementation for loading/storing vectors from/to memory segments.

Some of the changes are related to `int` -> `long` for the offset. I had to change the load/store intrinsic signatures to accept an offset as a `long`. This does not affect the C2 implementation, only the fallback lambdas. So in some cases there is a downcast to `int`, which is safe because the original input was an `int` e.g. the offset for primitive array access. 

Some changes related to module dependency and memory segment proxies will go away when the Memory Segment API is integrated into `java.base` as a preview API.
 
There is still more work to do but i wanted to get this in now and iterate:
- I will remove the `byte[]` and `ByteBuffer` accepting load/store methods in a following PR.
- some additional negative tests are required, specifically if a scope of segment is not alive, or the segment covers a heap array other than `byte[]`.

---------
### Progress

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 98eae994f8db06d9b2966e95a93abdb4d032f484

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/187/head:pull/187` \
`$ git checkout pull/187`

`$ git checkout pull/187` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/187/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 187`

`$ git pr show -t 187`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/187.diff">https://git.openjdk.java.net/panama-vector/pull/187.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorSpecies.java
+++ b/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorSpecies.java
@@ -149,7 +149,7 @@ public interface VectorSpecies<E> {
      * @return the largest multiple of the vector length not greater
      *         than the given length
      * @throws IllegalArgumentException if the {@code length} is
-               negative and the result would overflow to a positive value
+     *         negative and the result would overflow to a positive value
      * @see Math#floorMod(int, int)
      */
     int loopBound(int length);
@@ -174,7 +174,7 @@ public interface VectorSpecies<E> {
      * @return the largest multiple of the vector length not greater
      *         than the given length
      * @throws IllegalArgumentException if the {@code length} is
-    negative and the result would overflow to a positive value
+     *         negative and the result would overflow to a positive value
      * @see Math#floorMod(long, int)
      */
     long loopBound(long length);


merykitty:
Hi, may I ask why do we reject loading a vector from a heap `MemorySegment` not backed by a `byte[]`?

Thank.

PaulSandoz:
> Hi, may I ask why do we reject loading a vector from a heap `MemorySegment` not backed by a `byte[]`?
> 

I wanted to first keep to the equivalent functionality that is currently supported, then think through the implications of supporting any primitive backing array and what that means regarding alignment and byte order.

MemorySegment now supports misaligned access on primitive arrays for values whose size is larger or smaller than the primitive array element size. However, by default the value layouts disallow that, for example (using the preview API for memory segments), for example:

```java
MemorySegment ias = MemorySegment.ofArray(
        new int[] {0, 1, 2, 3, 4, 5, 6, 7, 8, 9});


// Access using element whose natural alignment is wider than the natural primitive array alignment

// java.lang.IllegalArgumentException: Misaligned access at address: 16
ias.set(JAVA_LONG, 0, 1L);

// No exception
ias.set(JAVA_LONG.withBitAlignment(8), 0, 1L);

// No exception
ias.set(JAVA_LONG.withBitAlignment(8), 1, 1L);


// Access using element whose natural alignment is smaller than the natural primitive array alignment

// java.lang.IllegalArgumentException: Misaligned access at address: 17
ias.set(JAVA_SHORT, 1, (short) 1);

// No exception
ias.set(JAVA_SHORT, 0, (short) 1);

// No exception
ias.set(JAVA_SHORT.withBitAlignment(8), 1, (short) 1);
```

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/ByteVector.java
```diff
@@ -3325,6 +3328,97 @@ ByteVector fromByteBuffer(VectorSpecies<Byte> species,
...
+                                           MemorySegment ms, long offset,
+                                           ByteOrder bo) {
+        offset = checkFromIndexSize(offset, species.vectorByteSize(), ms.byteSize());
+        ByteSpecies vsp = (ByteSpecies) species;
+        return vsp.dummyVector().fromMemorySegment0(ms, offset).maybeSwap(bo);
+    }
+
+    /**
+     * Loads a vector from a {@linkplain ByteBuffer byte buffer}
+     * starting at an offset into the byte buffer
```
XiaohongGong: Should be `{@linkplain MemorySegment memory segment}` and the `memory segment buffer` ?

XiaohongGong: Please ignore my comments since it is fixed in the second commit. Thanks!

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
```diff
@@ -550,6 +550,191 @@ public class ScopedMemoryAccess {
...
+        } catch (ScopedMemoryAccess.Scope.ScopedAccessError ex) {
+            throw new IllegalStateException("This segment is already closed");
+        }
+    }
+
+    @Scoped
+    @ForceInline
+    private static
+    <V extends VectorSupport.Vector<E>, E>
+    void storeIntoMemorySegmentScoped(ScopedMemoryAccess.Scope scope,
```
mcimadamore, line 674: I believe the rest of the class uses "internal" as a suffix for the scoped method

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Vector.java
```diff
@@ -3510,6 +3512,92 @@ public abstract void intoByteArray(byte[] a, int offset,
...
+     *         or {@code offset+(N+1)*ESIZE > ms.byteSize()}
+     *         for any lane {@code N} in the vector
+     * @throws UnsupportedOperationException
+     *         if the memory segment is read-only
+     * @throws IllegalArgumentException if the memory segment is a heap segment that is
+     *         not backed by a {@code byte[]} array.
+     * @throws IllegalStateException if the memory segment's session is not alive,
+     *         or if access occurs from a thread other than the thread owning the session.
+     */
+    public abstract void intoMemorySegment(MemorySegment ms, long offset, ByteOrder bo);
```
mcimadamore, line 3546: Just a note - in the memory segment API we used to have API points taking an explicit byte order (e.g. dereference methods) and it turned out these were just specific instances of a more general API method accepting a full layout. I sense the same might apply here - e.g. if you take a segment, an offset and a layout, then you might be able to generalize the API to more than byte[]-backed memory segment. But for now it's fine. Or - maybe, in this case the vector instance should have a lane layout somewhere, which can be used here.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Vector.java
```diff
@@ -3510,6 +3512,92 @@ public abstract void intoByteArray(byte[] a, int offset,
...
+     * <p>
+     * Bytes are extracted from primitive lane elements according
+     * to the specified byte ordering.
+     * The lanes are stored according to their
+     * <a href="Vector.html#lane-order">memory ordering</a>.
+     * <p>
+     * This method behaves as if it calls
+     * {@link #intoMemorySegment(MemorySegment,long,ByteOrder,VectorMask)
+     * intoMemorySegment()} as follows:
+     * <pre>{@code
```
mcimadamore, line 3527: you might want to revisit later to use code snippets

test/jdk/jdk/incubator/vector/Byte128VectorLoadStoreTests.java
```diff
@@ -729,6 +784,216 @@ static void loadStoreReadonlyByteBuffer(IntFunction<byte[]> fa,
...
+    }
+
+    @Test(dataProvider = "byteMemorySegmentProvider")
+    static void loadStoreReadonlyMemorySegment(IntFunction<byte[]> fa,
+                                               IntFunction<MemorySegment> fb,
+                                               ByteOrder bo) {
+        MemorySegment a = toSegment(fa.apply(SPECIES.length()), fb).asReadOnly();
+
+        try {
+            SPECIES.zero().intoMemorySegment(a, 0, bo);
```
mcimadamore, line 970: I think these can be written more succinctly with `expectThrows(UnsupportedOperationException.class, () -> { ... })`

mcimadamore (approved):
Looks good

PaulSandoz:
> @PaulSandoz I see, however, the problem of alignment here seems a little backward since a byte array is the most unaligned kind of array, so if we support loads, stores from byte arrays then alignment does not need any consideration anymore. 

Agreed for `byte[]` arrays. There is not much precedent right now for writing to other primitive arrays, whose element size differs from the vector lane element size. (Not an issue for the Vector API, but alignment does matters for a `byte[]` array when performing atomic access, since we cannot guarantee stable alignment of `byte[]` address across all platforms).

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
PaulSandoz: Done.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Vector.java
PaulSandoz: Yes, that's a good point, we need to revisit all of the doc.

test/jdk/jdk/incubator/vector/Byte128VectorLoadStoreTests.java
PaulSandoz: Done.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Vector.java
PaulSandoz: I need to think more thoroughly about this. The vector/species can have a layout (perhaps two, a value for the bits, and a sequence for lane element values), and the segment has a value layout for its elements, and the segment may be backed by a primitive array that also has an element value layout. Further complicating matters is the offset will be in units of byte.
(Separately it got me thinking if a segment should expose its "source" elemental layout, for native it would be `JAVA_BYTE`, for a heap covering `int[]` it would be `JAVA_INT`.)

mcimadamore (approved):
Looks good

rsmogura:
Very minor, I wonder if we should adjust `loopBound` to long sizes, too (I don't see it in diff)?

PaulSandoz:
> Very minor, I wonder if we should adjust `loopBound` to long sizes, too (I don't see it in diff)?

Good point. I will include that.

PaulSandoz:
@rsmogura just pushed an update.

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
```diff
@@ -580,11 +580,11 @@ public class ScopedMemoryAccess {
...
-    V loadFromMemorySegmentScoped(ScopedMemoryAccess.Scope scope,
-                                  Class<? extends V> vmClass, Class<E> e, int length,
-                                  MemorySegmentProxy msp, long offset,
-                                  S s,
-                                  VectorSupport.LoadOperation<MemorySegmentProxy, V, S> defaultImpl) {
+    V loadFromMemorySegmentScopedInternal(ScopedMemoryAccess.Scope scope,
+                                          Class<? extends V> vmClass, Class<E> e, int length,
+                                          MemorySegmentProxy msp, long offset,
+                                          S s,
+                                          VectorSupport.LoadOperation<MemorySegmentProxy, V, S> defaultImpl) {
```
XiaohongGong, line 587: Is it possible to merge this internal method with `loadFromByteBufferScopedInternal` ? And the same to other relative masked and store operations.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorSpecies.java
```diff
@@ -154,6 +154,31 @@
...
+     * {@link Math#floorMod(long,int) floorMod} computes a remainder
+     * value by rounding its quotient toward negative infinity.
+     * As long as {@code VLENGTH} is a power of two, then the result
+     * is also equal to {@code length & ~(VLENGTH - 1)}.
+     *
+     * @param length the input length
+     * @return the largest multiple of the vector length not greater
+     *         than the given length
+     * @throws IllegalArgumentException if the {@code length} is
+    negative and the result would overflow to a positive value
```
rsmogura, line 177: Just spotted this, small misprint here, no `*` in comment.

PaulSandoz: Thanks. Fixed.

src/java.base/share/classes/jdk/internal/misc/X-ScopedMemoryAccess.java.template
XiaohongGong: OK,  that's fine for me. Thanks!

PR #189, opened 2022-04-07
8284459: Add x86 back-end implementation for LEADING and TRAILING ZEROS COUNT operations

jatin-bhateja:
Summary of changes:
- Patch extends auto-vectorize to vectorize following Java SE APIs.
     1) Integer.numberOfLeadingZeros()
     2) Long.numberOfLeadingZeros()
     3) Integer.numberOfTrailingZeros()
     4) Long.numberOfTrailingZeros()

- Adds optimized X86 backend implementation for VectorOperations.LEADING_ZERO_COUNT and VectorOperations.TRAILING_ZEROS_COUNT for AVX512 and legacy targets.

Kindly review and share your feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8284459](https://bugs.openjdk.java.net/browse/JDK-8284459): Add x86 back-end implementation for LEADING and TRAILING ZEROS COUNT operations

### Reviewers
 * [Eric Liu](https://openjdk.java.net/census#eliu) (@theRealELiu - Committer) ⚠️ Review applies to cc42fb7c0dc8ea9d47d63c880417a6df4a28b60c

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/189/head:pull/189` \
`$ git checkout pull/189`

`$ git checkout pull/189` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/189/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 189`

`$ git pr show -t 189`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/189.diff">https://git.openjdk.java.net/panama-vector/pull/189.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4509,7 +4509,7 @@ void C2_MacroAssembler::vbroadcast(BasicType bt, XMMRegister dst, int imm32, Reg
   bool is_LP64 = LP64_ONLY(true) NOT_LP64(false);
   if ((is_LP64 || lane_size < 8) &&
       ((is_non_subword_integral_type(bt) && VM_Version::supports_avx512vl()) ||
-       (is_subword_type(bt) && VM_Version::supports_avx512bw()))) {
+       (is_subword_type(bt) && VM_Version::supports_avx512vlbw()))) {
     movptr(rtmp, imm32);
     switch(lane_size) {
       case 1 : evpbroadcastb(dst, rtmp, vec_enc); break;
@@ -4823,9 +4823,9 @@ void C2_MacroAssembler::vector_count_leading_zeros_evex(BasicType bt, XMMRegiste
     case T_SHORT:
       vpternlogd(xtmp1, 0xff, xtmp1, xtmp1, vec_enc);
       vpunpcklwd(xtmp2, xtmp1, src, vec_enc);
-      evplzcntd(xtmp2, k0, xtmp2, merge, vec_enc);
+      evplzcntd(xtmp2, ktmp, xtmp2, merge, vec_enc);
       vpunpckhwd(dst, xtmp1, src, vec_enc);
-      evplzcntd(dst, k0, dst, merge, vec_enc);
+      evplzcntd(dst, ktmp, dst, merge, vec_enc);
       vpackusdw(dst, xtmp2, dst, vec_enc);
       break;
     case T_BYTE:
@@ -4836,8 +4836,7 @@ void C2_MacroAssembler::vector_count_leading_zeros_evex(BasicType bt, XMMRegiste
       // Add T1 to T2 if 4 MSB bits of byte are all zeros.
       assert(VM_Version::supports_avx512bw(), "");
       evmovdquq(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()), vec_enc, rtmp);
-      movl(rtmp, 0x0F0F0F0F);
-      evpbroadcastd(dst, rtmp, vec_enc);
+      vbroadcast(T_INT, dst, 0x0F0F0F0F, rtmp, vec_enc);
       vpand(xtmp2, dst, src, vec_enc);
       vpshufb(xtmp2, xtmp1, xtmp2, vec_enc);
       vpsrlw(xtmp3, src, 4, vec_enc);
@@ -4854,10 +4853,8 @@ void C2_MacroAssembler::vector_count_leading_zeros_evex(BasicType bt, XMMRegiste
 
 void C2_MacroAssembler::vector_count_leading_zeros_byte_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
                                                             XMMRegister xtmp2, XMMRegister xtmp3, Register rtmp, int vec_enc) {
-  vmovdqu(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()));
-  movl(rtmp, 0x0F0F0F0F);
-  movdl(xtmp2, rtmp);
-  vpbroadcastd(xtmp2, xtmp2, vec_enc);
+  vmovdqu(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()), rtmp);
+  vbroadcast(T_INT, xtmp2, 0x0F0F0F0F, rtmp, vec_enc);
   // T1 = Compute leading zero counts of 4 LSB bits of each byte by
   // accessing the lookup table.
   vpand(dst, xtmp2, src, vec_enc);
@@ -4876,62 +4873,63 @@ void C2_MacroAssembler::vector_count_leading_zeros_byte_avx(XMMRegister dst, XMM
 
 void C2_MacroAssembler::vector_count_leading_zeros_short_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
                                                              XMMRegister xtmp2, XMMRegister xtmp3, Register rtmp, int vec_enc) {
-   vector_count_leading_zeros_byte_avx(dst, src, xtmp1, xtmp2, xtmp3, rtmp, vec_enc);
-   // Add zero counts of lower byte and upper byte of a word if
-   // upper byte holds a zero value.
-   vpsrlw(xtmp3, src, 8, vec_enc);
-   vpcmpeqw(xtmp3, xtmp1, xtmp3, vec_enc);
-   vpsllw(xtmp2, dst, 8, vec_enc);
-   vpaddw(xtmp2, xtmp2, dst, vec_enc);
-   vpblendvb(dst, dst, xtmp2, xtmp3, vec_enc);
-   vpsrlw(dst, dst, 8, vec_enc);
+  vector_count_leading_zeros_byte_avx(dst, src, xtmp1, xtmp2, xtmp3, rtmp, vec_enc);
+  // Add zero counts of lower byte and upper byte of a word if
+  // upper byte holds a zero value.
+  vpsrlw(xtmp3, src, 8, vec_enc);
+  // xtmp1 is set to all zeros by vector_count_leading_zeros_byte_avx.
+  vpcmpeqw(xtmp3, xtmp1, xtmp3, vec_enc);
+  vpsllw(xtmp2, dst, 8, vec_enc);
+  vpaddw(xtmp2, xtmp2, dst, vec_enc);
+  vpblendvb(dst, dst, xtmp2, xtmp3, vec_enc);
+  vpsrlw(dst, dst, 8, vec_enc);
 }
 
 void C2_MacroAssembler::vector_count_leading_zeros_int_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
                                                            XMMRegister xtmp2, XMMRegister xtmp3, int vec_enc) {
-   // Since IEEE 754 floating point format represents mantissa in 1.0 format
-   // hence biased exponent can be used to compute leading zero count as per
-   // following formula:-
-   // LZCNT = 32 - (biased_exp - 127)
-   // Special handling has been introduced for Zero, Max_Int and -ve source values.
+  // Since IEEE 754 floating point format represents mantissa in 1.0 format
+  // hence biased exponent can be used to compute leading zero count as per
+  // following formula:-
+  // LZCNT = 32 - (biased_exp - 127)
+  // Special handling has been introduced for Zero, Max_Int and -ve source values.
 
-   // Broadcast 0xFF
-   vpcmpeqd(xtmp1, xtmp1, xtmp1, vec_enc);
-   vpsrld(xtmp1, xtmp1, 24, vec_enc);
+  // Broadcast 0xFF
+  vpcmpeqd(xtmp1, xtmp1, xtmp1, vec_enc);
+  vpsrld(xtmp1, xtmp1, 24, vec_enc);
 
-   // Extract biased exponent.
-   vcvtdq2ps(dst, src, vec_enc);
-   vpsrld(dst, dst, 23, vec_enc);
-   vpand(dst, dst, xtmp1, vec_enc);
+  // Extract biased exponent.
+  vcvtdq2ps(dst, src, vec_enc);
+  vpsrld(dst, dst, 23, vec_enc);
+  vpand(dst, dst, xtmp1, vec_enc);
 
-   // Broadcast 127.
-   vpsrld(xtmp1, xtmp1, 1, vec_enc);
-   // Exponent = biased_exp - 127
-   vpsubd(dst, dst, xtmp1, vec_enc);
+  // Broadcast 127.
+  vpsrld(xtmp1, xtmp1, 1, vec_enc);
+  // Exponent = biased_exp - 127
+  vpsubd(dst, dst, xtmp1, vec_enc);
 
-   // Exponent = Exponent  + 1
-   vpsrld(xtmp3, xtmp1, 6, vec_enc);
-   vpaddd(dst, dst, xtmp3, vec_enc);
+  // Exponent = Exponent  + 1
+  vpsrld(xtmp3, xtmp1, 6, vec_enc);
+  vpaddd(dst, dst, xtmp3, vec_enc);
 
-   // Replace -ve exponent with zero, exponent is -ve when src
-   // lane contains a zero value.
-   vpxor(xtmp2, xtmp2, xtmp2, vec_enc);
-   vblendvps(dst, dst, xtmp2, dst, vec_enc);
+  // Replace -ve exponent with zero, exponent is -ve when src
+  // lane contains a zero value.
+  vpxor(xtmp2, xtmp2, xtmp2, vec_enc);
+  vblendvps(dst, dst, xtmp2, dst, vec_enc);
 
-   // Rematerialize broadcast 32.
-   vpslld(xtmp1, xtmp3, 5, vec_enc);
-   // Exponent is 32 if corresponding source lane contains max_int value.
-   vpcmpeqd(xtmp2, dst, xtmp1, vec_enc);
-   // LZCNT = 32 - exponent
-   vpsubd(dst, xtmp1, dst, vec_enc);
+  // Rematerialize broadcast 32.
+  vpslld(xtmp1, xtmp3, 5, vec_enc);
+  // Exponent is 32 if corresponding source lane contains max_int value.
+  vpcmpeqd(xtmp2, dst, xtmp1, vec_enc);
+  // LZCNT = 32 - exponent
+  vpsubd(dst, xtmp1, dst, vec_enc);
 
-   // Replace LZCNT with a value 1 if corresponding source lane
-   // contains max_int value.
-   vpblendvb(dst, dst, xtmp3, xtmp2, vec_enc);
+  // Replace LZCNT with a value 1 if corresponding source lane
+  // contains max_int value.
+  vpblendvb(dst, dst, xtmp3, xtmp2, vec_enc);
 
-   // Replace biased_exp with 0 if source lane value is less than zero.
-   vpxor(xtmp2, xtmp2, xtmp2, vec_enc);
-   vblendvps(dst, dst, xtmp2, src, vec_enc);
+  // Replace biased_exp with 0 if source lane value is less than zero.
+  vpxor(xtmp2, xtmp2, xtmp2, vec_enc);
+  vblendvps(dst, dst, xtmp2, src, vec_enc);
 }
 
 void C2_MacroAssembler::vector_count_leading_zeros_long_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
@@ -4940,6 +4938,7 @@ void C2_MacroAssembler::vector_count_leading_zeros_long_avx(XMMRegister dst, XMM
   // Add zero counts of lower word and upper word of a double word if
   // upper word holds a zero value.
   vpsrld(xtmp3, src, 16, vec_enc);
+  // xtmp1 is set to all zeros by vector_count_leading_zeros_byte_avx.
   vpcmpeqd(xtmp3, xtmp1, xtmp3, vec_enc);
   vpslld(xtmp2, dst, 16, vec_enc);
   vpaddd(xtmp2, xtmp2, dst, vec_enc);
@@ -5024,7 +5023,6 @@ void C2_MacroAssembler::vector_count_trailing_zeros_evex(BasicType bt, XMMRegist
                                                          XMMRegister xtmp1, XMMRegister xtmp2, XMMRegister xtmp3,
                                                          XMMRegister xtmp4, KRegister ktmp, Register rtmp, int vec_enc) {
   assert(is_integral_type(bt), "");
-  int bcast_value [] = { 0x8, 0x10, 0, 0x20, 0, 0, 0, 0x40 };
   // xtmp = -1
   vpternlogd(xtmp4, 0xff, xtmp4, xtmp4, vec_enc);
   // xtmp = xtmp + src
@@ -5032,7 +5030,7 @@ void C2_MacroAssembler::vector_count_trailing_zeros_evex(BasicType bt, XMMRegist
   // xtmp = xtmp & ~src
   vpternlogd(xtmp4, 0x40, xtmp4, src, vec_enc);
   vector_count_leading_zeros_evex(bt, dst, xtmp4, xtmp1, xtmp2, xtmp3, ktmp, rtmp, true, vec_enc);
-  vbroadcast(bt, xtmp4, bcast_value[type2aelembytes(bt) - 1], rtmp, vec_enc);
+  vbroadcast(bt, xtmp4, 8 * type2aelembytes(bt), rtmp, vec_enc);
   vpsub(bt, dst, xtmp4, dst, vec_enc);
 }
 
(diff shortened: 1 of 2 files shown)


merykitty:
Hi, for int vectors, I think you could convert it to float and look at the exponent part. For long vectors, the results can be obtained by counting each half and combined together. Thanks.

jatin-bhateja:
> Hi, for int vectors, I think you could convert it to float and look at the exponent part. For long vectors, the results can be obtained by counting each half and combined together. Thanks.

Hi, nice suggestion!, for almost all AVX512 targets (AVX512CD) we do have direct instruction to compute leading zero count for both integer and long vectors, in addition quad word to double precision conversion instruction is supported only for AVX512DQ targets. But I think we can still manage to generate efficient sequence for integer vectors for AVX2 target by extracting biased exponents.

src/hotspot/share/opto/loopTransform.cpp
```diff
@@ -975,6 +975,10 @@ bool IdealLoopTree::policy_unroll(PhaseIdealLoop *phase) {
         const TypeVect* vt = n->bottom_type()->is_vect();
         body_size += Matcher::vector_op_pre_select_sz_estimate(n->Opcode(), vt->element_basic_type(), vt->length());
       } break;
+      case Op_CountLeadingZerosV: {
+        const TypeVect* vt = n->bottom_type()->is_vect();
+        body_size += Matcher::vector_op_pre_select_sz_estimate(n->Opcode(), vt->element_basic_type(), vt->length());
+      } break;
```
theRealELiu, line 981: It can share the same code with Op_PopCountVI/L.

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -93,7 +93,7 @@ class VectorNode : public TypeNode {
   static bool is_type_transition_short_to_int(Node* n);
   static bool is_type_transition_to_int(Node* n);
```
theRealELiu, line 94: Not sure if we should add these 2 l2i candidates into `is_type_transition_to_int`. As they are indeed transformed to int.

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -93,7 +93,7 @@ class VectorNode : public TypeNode {
...
   static bool is_muladds2i(Node* n);
-  static bool is_vpopcnt_long(Node* n);
+  static bool is_downcasting_l2i_candidate(Node* n);
```
theRealELiu, line 96: I noticed there is  `static bool is_type_transition_short_to_int(Node* n)` above, how do you think about `is_type_transition_long_to_int`?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -1253,6 +1258,7 @@ public:
   };
 };
 
+
```
sviswa7, line 1261: Extra empty line.

src/hotspot/share/opto/loopTransform.cpp
```diff
@@ -975,6 +975,7 @@ bool IdealLoopTree::policy_unroll(PhaseIdealLoop *phase) {
         const TypeVect* vt = n->bottom_type()->is_vect();
         body_size += Matcher::vector_op_pre_select_sz_estimate(n->Opcode(), vt->element_basic_type(), vt->length());
       } break;
+      case Op_CountLeadingZerosV:
       case Op_ReverseV: {
```
sviswa7, line 979: These two cases could merge with the Op_PopCountVL. The body is the same.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4723,174 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+  vpcmpeqb(xtmp3, xtmp1, xtmp3, vec_enc);
+  vpaddb(dst, dst, xtmp2, vec_enc);
+  vpblendvb(dst, xtmp2, dst, xtmp3, vec_enc);
+}
+
+void C2_MacroAssembler::vector_count_leading_zeros_short_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
+                                                             XMMRegister xtmp2, XMMRegister xtmp3, Register rtmp, int vec_enc) {
+   vector_count_leading_zeros_byte_avx(dst, src, xtmp1, xtmp2, xtmp3, rtmp, vec_enc);
+   // Add zero counts of lower byte and upper byte of a word if
+   // upper byte holds a zero value.
```
sviswa7, line 4798: Need to specify here that xmm1 is set to all zeros by vector_count_leading_zeros_byte_avx.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4723,174 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+   vpsrlw(xtmp3, src, 8, vec_enc);
+   vpcmpeqw(xtmp3, xtmp1, xtmp3, vec_enc);
+   vpsllw(xtmp2, dst, 8, vec_enc);
+   vpaddw(xtmp2, xtmp2, dst, vec_enc);
+   vpblendvb(dst, dst, xtmp2, xtmp3, vec_enc);
```
sviswa7, line 4803: The mask is generated using a word operation, but blend is a byte operation?

jatin-bhateja: There is no variable word/double word/quadword blend,  generated mask lane is composed of multiple bytes and hence byte level blend is sufficient to blend multi-byte lanes. We do have a single/double precision blend but its latency is similar to byte level blend in addition it may incur domain switch over penalty.

theRealELiu (approved):
Shared code looks good to me.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4433,14 +4433,30 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
...
-void C2_MacroAssembler::vbroadcastd(XMMRegister dst, int imm32, Register rtmp, int vec_enc) {
-  if (VM_Version::supports_avx512vl()) {
-    movl(rtmp, imm32);
-    evpbroadcastd(dst, rtmp, vec_enc);
+void C2_MacroAssembler::vbroadcast(BasicType bt, XMMRegister dst, int imm32, Register rtmp, int vec_enc) {
+  int lane_size = type2aelembytes(bt);
+  bool is_LP64 = LP64_ONLY(true) NOT_LP64(false);
+  if ((is_LP64 || lane_size < 8) &&
+      ((is_non_subword_integral_type(bt) && VM_Version::supports_avx512vl()) ||
+       (is_subword_type(bt) && VM_Version::supports_avx512bw()))) {
```
sviswa7, line 4441: The vl check is needed for all lane sizes when vector width < 64 bytes. The check doesn't seem to capture that.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
sviswa7: This is still pending.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+  assert(is_integral_type(bt), "");
+  int bcast_value [] = { 0x8, 0x10, 0, 0x20, 0, 0, 0, 0x40 };
+  // xtmp = 0
+  vpxor(xtmp3 , xtmp3, xtmp3, vec_enc);
+  // xtmp = 0 - src
+  vpsub(bt, xtmp3, xtmp3, src, vec_enc);
+  // xtmp = xtmp | src
+  vpor(xtmp3, xtmp3, src, vec_enc);
+  vector_popcount_integral(bt, dst, xtmp3, xtmp1, xtmp2, rtmp, vec_enc);
+  vbroadcast(bt, xtmp1, bcast_value[type2aelembytes(bt) - 1], rtmp, vec_enc);
```
sviswa7, line 4981: The bcast_value could be replaced by (0x8 * type2aelembytes(bt)).

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+      movl(rtmp, 0x0F0F0F0F);
+      evpbroadcastd(dst, rtmp, vec_enc);
+      vpand(xtmp2, dst, src, vec_enc);
+      vpshufb(xtmp2, xtmp1, xtmp2, vec_enc);
+      vpsrlw(xtmp3, src, 4, vec_enc);
+      vpand(xtmp3, dst, xtmp3, vec_enc);
+      vpshufb(dst, xtmp1, xtmp3, vec_enc);
+      vpxor(xtmp1, xtmp1, xtmp1, vec_enc);
+      evpcmpeqb(ktmp, xtmp1, xtmp3, vec_enc);
+      evpaddb(dst, ktmp, dst, xtmp2, true, vec_enc);
```
sviswa7, line 4777: It is possible to do this without needing xtmp3:
     // Nibble clz table in xtmp1
      evmovdquq(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()), vec_enc, rtmp);
     // Nibble mask in xtmp2 
      movl(rtmp, 0x0F0F0F0F);
      evpbroadcastd(xtmp2, rtmp, vec_enc);  
      // Get upper nibble in low 4 bits of dst
      vpsrlw(dst, src, 4, vec_enc);
      vpand(dst, dst, xtmp2, vec_enc);
      // Get clz of upper nibble into dst using table in xtmp1
      vpshufb(dst, xtmp1, dst, vec_enc);
      // Get lower nibble in low 4 bits of xtmp2 overwriting the nibble mask
      vpand(xtmp2, xtmp2, src, vec_enc);
      // Get clz of lower nibble in xtmp2 using the table in xtmp1
      vpshufb(xtmp2, xtmp1, xtmp2, vec_enc);
      // Broadcast the clz of 0 into all lanes of xtmp1, note the lowest byte had clz of zero in the xtmp1 table 
      evpbroadcastb(xtmp1, xtmp1, xtmp1, vec_enc);
      // Check if the clz of upper nibble in dst indicates that the upper nibble was all zero
      evpcmpeqb(ktmp, xtmp1, dst, vec_enc);
      // if upper nibble was zero add the clz of lower nibble to dst
      evpaddb(dst, ktmp, dst, xtmp2,  true, vec_enc);

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+                                                        KRegister ktmp, Register rtmp, bool merge, int vec_enc) {
+  assert(is_integral_type(bt), "");
+  assert(VM_Version::supports_avx512vl() || vec_enc == Assembler::AVX_512bit, "");
+  assert(VM_Version::supports_avx512cd(), "");
+  switch(bt) {
+    case T_LONG:
+      evplzcntq(dst, ktmp, src, merge, vec_enc);
+      break;
+    case T_INT:
+      evplzcntd(dst, ktmp, src, merge, vec_enc);
```
sviswa7, line 4750: The ktmp here should be k0. An assert here or use explicit k0.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+      evpaddb(dst, ktmp, dst, xtmp2, true, vec_enc);
+      break;
+    default:
+      ShouldNotReachHere();
+  }
+}
+
+void C2_MacroAssembler::vector_count_leading_zeros_byte_avx(XMMRegister dst, XMMRegister src, XMMRegister xtmp1,
+                                                            XMMRegister xtmp2, XMMRegister xtmp3, Register rtmp, int vec_enc) {
+  vmovdqu(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()));
```
merykitty, line 4786: Please pass `rtmp` here, thanks.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+  // T2 = Compute leading zero counts of 4 MSB bits of each byte by
+  // accessing the lookup table.
+  vpsrlw(xtmp3, src, 4, vec_enc);
+  vpand(xtmp3, xtmp2, xtmp3, vec_enc);
+  vpshufb(xtmp2, xtmp1, xtmp3, vec_enc);
+  // Add T1 to T2 if 4 MSB bits of byte are all zeros.
+  vpxor(xtmp1, xtmp1, xtmp1, vec_enc);
+  vpcmpeqb(xtmp3, xtmp1, xtmp3, vec_enc);
+  vpaddb(dst, dst, xtmp2, vec_enc);
+  vpblendvb(dst, xtmp2, dst, xtmp3, vec_enc);
```
sviswa7, line 4803: vector_count_leading_zeros_byte_avx can be implemented without xmpt3 as follows:
 // Nibble clz table in xtmp1
 vmovdqu(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()), rtmp);
 
// Nibble mask in xtmp2
 vbroadcast(T_INT, xtmp2, 0x0F0F0F0F, rtmp, vec_enc);

  // dst = Compute leading zero counts of 4 MSB bits of each byte by
  // accessing the lookup table
  vpsrlw(dst, src, 4, vec_enc);
  vpand(dst, dst, xtmp2, vec_enc);
  vpshufb(dst, xtmp1, dst, vec_enc);

  // xtmp2 = Compute leading zero counts of 4 LSB bits of each byte by
  // accessing the lookup table.
  vpand(xtmp2, xtmp2, src, vec_enc);
  vpshufb(xtmp2, xtmp1, xtmp2, vec_enc);
  
  // Add xtmp2 to dst if 4 MSB bits of byte are all zeros i.e. if the dst had clz of 0
  vpbroadcastb(xtmp1, xtmp1, vec_enc); 
  vpcmpeqb(xtmp1, xtmp1, dst, vec_enc);
  vpaddb(xtmp2, xtmp2, dst, vec_enc);
  vpblendvb(dst, dst, xtmp2, xtmp1, vec_enc);

src/hotspot/cpu/x86/stubGenerator_x86_32.cpp
```diff
@@ -588,6 +588,30 @@ class StubGenerator: public StubCodeGenerator {
...
+    __ emit_data(0x00000000, relocInfo::none, 0);
+    __ emit_data(0x00000000, relocInfo::none, 0);
+    __ emit_data(0x02020304, relocInfo::none, 0);
+    __ emit_data(0x01010101, relocInfo::none, 0);
+    __ emit_data(0x00000000, relocInfo::none, 0);
+    __ emit_data(0x00000000, relocInfo::none, 0);
+    __ emit_data(0x02020304, relocInfo::none, 0);
+    __ emit_data(0x01010101, relocInfo::none, 0);
+    __ emit_data(0x00000000, relocInfo::none, 0);
+    __ emit_data(0x00000000, relocInfo::none, 0);
```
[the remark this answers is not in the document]
jatin-bhateja: PSHUFB operates at a lane level (16 bytes), thus in order to perform lookup for upper lane we either load 32/64 bytes from memory or we will have to replicated lower lane across vector. I choose first approach to be consistent with other LUT based implementations.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -795,6 +795,21 @@ class StubGenerator: public StubCodeGenerator {
...
+    StubCodeMark mark(this, "StubRoutines", stub_name);
+    address start = __ pc();
+    __ emit_data64(0x0101010102020304, relocInfo::none);
+    __ emit_data64(0x0000000000000000, relocInfo::none);
+    __ emit_data64(0x0101010102020304, relocInfo::none);
+    __ emit_data64(0x0000000000000000, relocInfo::none);
+    __ emit_data64(0x0101010102020304, relocInfo::none);
+    __ emit_data64(0x0000000000000000, relocInfo::none);
+    __ emit_data64(0x0101010102020304, relocInfo::none);
+    __ emit_data64(0x0000000000000000, relocInfo::none);
```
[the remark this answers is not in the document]
jatin-bhateja: Same as above.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8669,35 +8684,107 @@ instruct vpopcount_integral_reg_evex_masked(vec dst, vec src, kReg mask) %{
   ins_pipe( pipe_slow );
 %}
 
-instruct vpopcountI_avx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) %{
-  predicate((!VM_Version::supports_avx512_vpopcntdq() && Matcher::vector_element_basic_type(n->in(1)) == T_INT) ||
-            (!VM_Version::supports_avx512_bitalg() && is_subword_type(Matcher::vector_element_basic_type(n->in(1)))));
+instruct vpopcount_evx_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegP rtmp) %{
```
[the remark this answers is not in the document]
jatin-bhateja: Corrected.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -8669,35 +8684,107 @@ instruct vpopcount_integral_reg_evex_masked(vec dst, vec src, kReg mask) %{
...
+  predicate(!is_vector_popcount_predicate(Matcher::vector_element_basic_type(n->in(1))));
   match(Set dst (PopCountVI src));
+  match(Set dst (PopCountVL src));
   effect(TEMP dst, TEMP xtmp1, TEMP xtmp2, TEMP rtmp);
-  format %{ "vector_popcount_int  $dst, $src\t! using $xtmp1, $xtmp2 and $rtmp as TEMP" %}
+  format %{ "vector_popcount_integral $dst, $src\t! using $xtmp1, $xtmp2, and $rtmp as TEMP" %}
   ins_encode %{
+    int opcode = this->ideal_Opcode();
+    int vlen_enc = vector_length_encoding(this, $src);
     BasicType bt = Matcher::vector_element_basic_type(this);
```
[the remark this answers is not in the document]
jatin-bhateja: Corrected, in other patterns I am extracting basic type from src.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: DONE

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+    case T_BYTE:
+      // T1 = Compute leading zero counts of 4 LSB bits of each byte by
+      // accessing the lookup table.
+      // T2 = Compute leading zero counts of 4 MSB bits of each byte by
+      // accessing the lookup table.
+      // Add T1 to T2 if 4 MSB bits of byte are all zeros.
+      assert(VM_Version::supports_avx512bw(), "");
+      evmovdquq(xtmp1, ExternalAddress(StubRoutines::x86::vector_count_leading_zeros_lut()), vec_enc, rtmp);
+      movl(rtmp, 0x0F0F0F0F);
+      evpbroadcastd(dst, rtmp, vec_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: In case of noreg assembler will fail due to assertion failure.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+      break;
+    case T_INT:
+      evplzcntd(dst, ktmp, src, merge, vec_enc);
+      break;
+    case T_SHORT:
+      vpternlogd(xtmp1, 0xff, xtmp1, xtmp1, vec_enc);
+      vpunpcklwd(xtmp2, xtmp1, src, vec_enc);
+      evplzcntd(xtmp2, k0, xtmp2, merge, vec_enc);
+      vpunpckhwd(dst, xtmp1, src, vec_enc);
+      evplzcntd(dst, k0, dst, merge, vec_enc);
```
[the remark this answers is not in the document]
jatin-bhateja: This is a common routine which gets used for both predicated (in case of INT/LONG) and non-predicated count zeros, Replacing k0 with ktmp and passing k0 from top level patterns will add the needed consistency.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4723,3 +4735,249 @@ void C2_MacroAssembler::vector_reverse_byte(BasicType bt, XMMRegister dst, XMMRe
...
+  assert(VM_Version::supports_avx512vl() || vec_enc == Assembler::AVX_512bit, "");
+  assert(VM_Version::supports_avx512cd(), "");
+  switch(bt) {
+    case T_LONG:
+      evplzcntq(dst, ktmp, src, merge, vec_enc);
+      break;
+    case T_INT:
+      evplzcntd(dst, ktmp, src, merge, vec_enc);
+      break;
+    case T_SHORT:
```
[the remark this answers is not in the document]
jatin-bhateja: xnoreg represents -1 value, which results into an assertion failure due to illegal register encoding.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: Thanks, Sandhya,  you have introduced additional broadcast instruction to save a temporary.  Spill to memory may be costly too but adding an explicit instruction in anticipation of a spill may hit the performance. Also we will not be able to further merge this instruction pattern even after removing additional temporary.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: ktmp will be a non k0 opmask register for predicated leading zero count only for INT/LONG type.   match_rule_supported_vector_masked carries necessary conditions.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: Thanks, Sandhya,  you have introduced additional broadcast instruction to save a temporary.  Spill to memory may be costly too but adding an explicit instruction in anticipation of a spill may hit the performance. Also we will not be able to further merge this instruction pattern even after removing additional temporary. 

// Add xtmp2 to dst if 4 MSB bits of byte are all zeros i.e. if the dst had clz of 0
We add LZ count of 4MSB bits if they are all zeros i.e. clz of 4.

jatin-bhateja:
Hi @sviswa7 , your comments have been addressed.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -9127,6 +9215,112 @@ instruct vreverse_byte64_reg(vec dst, vec src, vec xtmp1, vec xtmp2, rRegI rtmp)
...
+  predicate(is_clz_non_subword_predicate_evex(Matcher::vector_element_basic_type(n->in(1)),
+                                              Matcher::vector_length_in_bytes(n->in(1))));
+  match(Set dst (CountLeadingZerosV src mask));
+  format %{ "vector_count_leading_zeros $dst, $src, $mask" %}
+  ins_encode %{
+    int vlen_enc = vector_length_encoding(this, $src);
+    BasicType bt = Matcher::vector_element_basic_type(this, $src);
+    __ evmovdquq($dst$$XMMRegister, $src$$XMMRegister, vlen_enc);
+    __ vector_count_leading_zeros_evex(bt, $dst$$XMMRegister, $src$$XMMRegister, xnoreg, xnoreg,
+                                       xnoreg, $mask$$KRegister, noreg, true, vlen_enc);
```
sviswa7, line 9251: For the PopCountVL and CountLeadingZerosV for long, if the result type is T_INT the evpmovqd instruction needs to be generated.

sviswa7 (approved):
I only have one comment remaining. We do auto-vectorize PopCountVL today. The masked support for auto-vectorized tail loop was added recently on mainline. So on masked path we should handle the conversion from long to int when the result type is int vector.  Rest of the patch looks good to me. Please fix and integrate.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: Its only through vector API will land in this control path and in case of vector API result lane size matches the source lane size.

jatin-bhateja:
> I only have one comment remaining. We do auto-vectorize PopCountVL today. The masked support for auto-vectorized tail loop was added recently on mainline. So on masked path we should handle the conversion from long to int when the result type is int vector. Rest of the patch looks good to me. Please fix and integrate.

Hi @sviswa7 , Thanks for your comments, post-loop tail vectorization support does not generate predicated operations apart from vector load/store currently, since SLP algorithm operates over expression trees delimited by memory operations [or end in scalarized users] hence just predicating load/store is sufficient to ensure correct semantics.

PR #190, opened 2022-04-07
Remove `byte[]` and `ByteBuffer` access methods.

PaulSandoz:
Removes `byte[]` and `ByteBuffer` vector access methods.

Fixes performance issue with the fallback implementation of `MemorySegment` accessors (the `ValueLayout` passed to the scalar accessors needs to be constant).

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-vector pull/190/head:pull/190` \
`$ git checkout pull/190`

`$ git checkout pull/190` \
`$ git pull https://git.openjdk.java.net/panama-vector pull/190/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 190`

`$ git pr show -t 190`

</details>

<a href="https://git.openjdk.java.net/panama-vector/pull/190.diff">https://git.openjdk.java.net/panama-vector/pull/190.diff</a>

</details>


DamonFool:
Hi @PaulSandoz ,

Is there any perf number before and after this patch?
Also why there is no JBS for this PR?

PaulSandoz:
Hi @DamonFool, issues are optional for integration into the `panama-vector` repo.

Performance-wise `MemorySegment` is similar to `byte[]` and `ByteBuffer` (heap and direct), which is not too surprising since the `base`/`offset` are extracted from a segment and passed to the `load`/`store` vector intrinsic.

There is one performance anomaly I have observed related to using an `int`-based loop and `int` offset. This results in a widening conversion to `long` when passing the `int` value as the `offset` argument to the access method. The memory addressing logic is not efficient, which exacerbates with unrolling. This may be more generally C2 related, but it requires further investigation.

PaulSandoz:
@merykitty that's a good point. I have run into that before, and could be a likely cause.

DamonFool:
> Hi @DamonFool, issues are optional for integration into the `panama-vector` repo.

Okay.

> 
> Performance-wise `MemorySegment` is similar to `byte[]` and `ByteBuffer` (heap and direct), which is not too surprising since the `base`/`offset` are extracted from a segment and passed to the `load`/`store` vector intrinsic.
> 
> There is one performance anomaly I have observed related to using an `int`-based loop and `int` offset. This results in a widening conversion to `long` when passing the `int` value as the `offset` argument to the access method. The memory addressing logic is not efficient, which exacerbates with unrolling. This may be more generally C2 related, but it requires further investigation.

Can you give us a benchmark example whose performance would be improved so that we can learn more from your patch?
Thanks.

PaulSandoz:
@DamonFool I confirmed what @merykitty described, related to overflow, although I suppose if the loop bound is known then it could be determined if overflow occurs or not. Although, as i understand this area of HotSpot is quite tricky.

If doing a calculation such as `someIntIndex << 2` it is necessary to explicitly cast `someIntIndex` to a `long`, similar to what is required when using `Unsafe` e.g. see [here](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java#L760). The alternative is to use a `long` based index.

sviswa7 (approved):
Looks good to me.

PR #201, opened 2022-05-25
8287289: Gather/Scatter with Index Vector

JoshuaZhuwj:
When I assist engineers to apply VectorAPI in real business scenarios, I realize that Gather/Scatter APIs depend on indexMap residing in memory.

When an index map is acquired by vector operations, it is represented by IntVector. To do Gather/Scatter operations, an extra integer array must be introduced and an explicit vector store is also required ahead of each Gather/Scatter. Furthermore, the redundant memory store may cause a performance penalty.

Hence I submit this change for discussion. I propose to provide Gather/Scatter API supporting index vector. This patch only includes the change for Gather API.
It passed the jtreg tests for VectorAPI.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8287289](https://bugs.openjdk.org/browse/JDK-8287289): Gather/Scatter with Index Vector (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-vector.git pull/201/head:pull/201` \
`$ git checkout pull/201`

`$ git checkout pull/201` \
`$ git pull https://git.openjdk.org/panama-vector.git pull/201/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 201`

`$ git pr show -t 201`

</details>

<a href="https://git.openjdk.org/panama-vector/pull/201.diff">https://git.openjdk.org/panama-vector/pull/201.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-vector/pull/201#issuecomment-1136964353)

### DIFF
--- a/src/hotspot/share/classfile/vmIntrinsics.hpp
+++ b/src/hotspot/share/classfile/vmIntrinsics.hpp
@@ -1080,11 +1080,10 @@ class methodHandle;
                                      "Ljdk/internal/vm/vector/VectorSupport$Vector;"                                                           \
                                      "Ljdk/internal/vm/vector/VectorSupport$VectorMask;"                                                       \
                                      "Ljava/lang/Object;"                                                                                      \
-                                     "I[II"                                                                                                    \
                                      "Ljdk/internal/vm/vector/VectorSupport$VectorSpecies;"                                                    \
-                                     "Ljdk/internal/vm/vector/VectorSupport$LoadVectorOperationWithMap;)"                                      \
+                                     "Ljdk/internal/vm/vector/VectorSupport$LoadVectorOperationWithIndexMap;)"                                 \
                                      "Ljdk/internal/vm/vector/VectorSupport$Vector;")                                                          \
-    do_name(vector_gather_name, "loadWithMap")                                                                                                 \
+    do_name(vector_gather_name, "loadWithIndexMap")                                                                                            \
                                                                                                                                                \
    do_intrinsic(_VectorScatterOp, jdk_internal_vm_vector_VectorSupport, vector_scatter_name, vector_scatter_sig, F_S)                          \
     do_signature(vector_scatter_sig, "(Ljava/lang/Class;"                                                                                      \
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -1361,12 +1361,12 @@ bool LibraryCallKit::inline_vector_mem_masked_operation(bool is_store) {
 //  S extends VectorSpecies<E>,
 //  M extends VectorMask<E>,
 //  E>
-// V loadWithMap(Class<? extends V> vectorClass, Class<M> maskClass, Class<E> elementType, int length,
-//               Class<? extends Vector<Integer>> vectorIndexClass,
-//               Object base, long offset, // Unsafe addressing
-//               W index_vector, M m,
-//               C container, int index, int[] indexMap, int indexM, S s, // Arguments for default implementation
-//               LoadVectorOperationWithMap<C, V, E, S, M> defaultImpl)
+//  V loadWithIndexMap(Class<? extends V> vClass, Class<M> mClass, Class<E> eClass, int length,
+//                     Class<? extends Vector<Integer>> vectorIndexClass,
+//                     Object base, long offset,
+//                     W index_vector, M m,
+//                     C container, S s,
+//                     LoadVectorOperationWithIndexMap<C, V, W, S, M> defaultImpl)
 //
 //  <C,
 //   V extends Vector<E>,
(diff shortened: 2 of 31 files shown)


JoshuaZhuwj:
Current Gather/Scatter APIs accept int array as index map. It hides the internal details of how the int array gets involved in vectorization, making users no need to pay attention to the underlying index vector's length.
But in some scenarios, the index map has to participate in the vector computation and then be obtained from the previous iteration. In this kind of scenario, Gather/Scatter with index vector will be much more helpful.

merykitty:
I think we could use `VectorShuffle` for the index map argument. They are an index vector, after all. A problem would be that `VectorShuffle` is backed up by a byte vector, which limits the indices that can be referred to by gather/scatter. But IMO we should address that problem also. Thanks.

JoshuaZhuwj:
> I think we could use `VectorShuffle` for the index map argument. They are both index vectors, after all.
Thanks for your comments. I think using "VectorShuffle" in Gather/Scatter to represent index map is a good suggestion.

>A problem would be that `VectorShuffle` is backed up by a byte vector, which limits the indices that can be referred to by gather/scatter. But IMO we should address that problem also. Thanks.
Yes. I'm thinking about how much influence will bring if replace byte vector with int vector.

JoshuaZhuwj:
> I think we could use `VectorShuffle` for the index map argument. They are both index vectors, after all.

Thanks for your comments. Agree, I think using "VectorShuffle" in Gather/Scatter to represent index map is a good suggestion.

> A problem would be that `VectorShuffle` is backed up by a byte vector, which limits the indices that can be referred to by gather/scatter. But IMO we should address that problem also. Thanks.

Yes. I'm thinking about how much influence will bring if replace byte vector with int vector.

merykitty:
> Yes. I'm thinking about how much influence will bring if replace byte vector with int vector.

I think it should be backed by a vector of the element type instead.

rose00:
P.S. As long as I'm talking about scatter and gather, I should also point out that the above APIs (as well as the existing proof-of-concept APIs) do not support scatter or gather to/from more than one Java object at a time.  This is another limitation I'd like to eventually remove.

An API which could do this looks like 2-D scatter/gathers where the first dimension is an index (a real `int`) into a Java array of multiple objects.  Panama lets us declare that this array can be of type `MemorySegment[]` and that's all you need.

```
FloatVector fromArrays(
   VectorSpecies<Float> species,
   MemorySegment[] as,
   IntVector indexVector1,
   IntVector indexVector2);
FloatVector fromArrays(
   VectorSpecies<Float> species,
   MemorySegment[] as,
   IntVector indexVector1,
   LongVector indexVector2);
// and for constant second index:
FloatVector fromArrays(
   VectorSpecies<Float> species,
   MemorySegment[] as,
   IntVector indexVector1,
   int index2);
FloatVector fromArrays(
   VectorSpecies<Float> species,
   MemorySegment[] as,
   IntVector indexVector1,
   long index2);
```

P.P.S. I suppose that the word `Array` in all of the above should be called `Memory` if embrace Panama.  It's not really an array, except in a logical sense.  Also I have omitted any discussion of index scaling and alignment.  There might be more to say about those issues.

JoshuaZhuwj:
John, thanks a lot for your comments.

> If we can figure out how to do scatter-gather over memory segments, that would support Java arrays (perhaps via convenience methods) and also support more general native programming scenarios, including ones which are not limited to 31-bit indexes into the Java heap.

Make sense, I will follow your suggestion and implement Gather/Scatter operation over memory segments by int/long index.
I think besides Gather/Scatter API on MemorySegment, the existing proof-of-concept APIs over Java arrays are still needed, right? As for the implementation of them, I will figure out whether the way that depends on memory segments is better.

> As a side note, I'd also like to see uniform support of subword types for scatter and gather. This is messy, but part of the value of Java is tastefully hiding such messes.

OK. Thanks.

JoshuaZhuwj:
John Rose, thanks for your reply:-)

Yes, based on ease-of-use considerations, more APIs have to be introduced to support all these choices.
In my opinion, letting users build these capabilities on top of the combination of fundamental APIs looks more elegant and reasonable if they could be optimized well by JIT.
Hence this requires the fundamental API could meet different requirements in diverse scenarios, such as removing the memory assumption for indexMap in this case.

As you suggest, as the first step, I will implement the fundamental Gather/Scatter API looks like this:

    LongVector fromMemorySegment(VectorSpecies<Long> species,
                                 MemorySegment ms,
                                 Vector<Integer> indexVector,
                                 ByteOrder bo);

    LongVector fromMemorySegment(VectorSpecies<Long> species,
                                 MemorySegment ms,
                                 Vector<Long> indexVector,
                                 ByteOrder bo);

I agree with your choice of offset-mode.
Non-scaled offsets are more suitable for index in memory segments version.
Support of scaled indexes could be extended base on it.
Thanks a lot for your suggestion on it.

If we decide to add more capabilities at API level, we could build them on top of fundamental APIs as you told us.

merykitty:
Hi, are you still working on this?

JoshuaZhuwj:
> Hi, are you still working on this?

I haven't worked on it yet. Please let me know if you're interested.

DirkToewe:
I would be interested. Index vector are much more flexible and might even give some performance benefits.

PR #213, opened 2023-04-07
8287289: Gather/Scatter with Index Vector

merykitty:
Hi,

This patch implements gather/scatter operations using an index vector in the form of an `IntVector` or a `LongVector` on arrays. According to the rationales in the [original PR](https://github.com/openjdk/panama-vector/pull/201), gathering/scattering using a vector is more flexible and can be used to implement the original interface of using an index array easily. Furthermore, 64-bit indexing is necessary for operating on `MemorySegment`s in the future, which can span a larger-than-4GiB space.

Please take a look and leave reviews.
Thanks a lot.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Commit message must refer to an issue

### Integration blockers
&nbsp;⚠️ Too few reviewers with at least role reviewer found (have 0, need at least 1) (failed with the updated jcheck configuration)
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Issue
 * [JDK-8287289](https://bugs.openjdk.org/browse/JDK-8287289): Gather/Scatter with Index Vector (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-vector.git pull/213/head:pull/213` \
`$ git checkout pull/213`

`$ git checkout pull/213` \
`$ git pull https://git.openjdk.org/panama-vector.git pull/213/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 213`

`$ git pr show -t 213`

</details>

<a href="https://git.openjdk.org/panama-vector/pull/213.diff">https://git.openjdk.org/panama-vector/pull/213.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-vector/pull/213#issuecomment-1500389645)


merykitty:
I am not confident in implementing the AArch64 port of this patch, so it would be appreciated if someone can help me with it, thanks very much in advance.

merykitty:
Another question is that where should I base this patch onto? I assume it should be `vectorIntrinsics`.

PaulSandoz:
> Another question is that where should I base this patch onto? I assume it should be `vectorIntrinsics`.

Since this is relatively small in scope I think it suitable as an enhancement for the jdk repo, and if ready could be targeted for 21. But, it may take us a few iterations to get the API right.

- Why did you remove mapped index access using MemorySegments? Its tempting to focus on that as the primitive, and not add specific API points for arrays, given these access methods are more specialized. Perhaps you ran into some issues? 

- Initially I was pondering whether we should also have an offset argument, but this only seems worthwhile if there are address calculation optimizations that cannot be easily applied (if any are possible) to the user code adding an offset to the input mapped index vector (also see next point).

- Having IntVector and LongVector for MemorySegments seems unavoidable given species lane counts. For arrays we could probably get away with just IntVector, but we could also just drop those too if MemorySegments worked out. However, for MemorySegments using IntVector we would need to point out that the segment might need to be sliced at a required offset.

test/hotspot/jtreg/compiler/vectorapi/VectorGatherScatterTest.java
```diff
@@ -42,7 +38,6 @@
  * @bug 8288397
  * @key randomness
  * @library /test/lib /
- * @requires vm.cpu.features ~= ".*sve.*" & (vm.opt.MaxVectorSize == "null" | vm.opt.MaxVectorSize >= 16)
```
XiaohongGong, line 45: We have to make sure this test is run with "MaxVectorSize >= 16", or following code will throw exception due to no vector shape maps (e.g. `32-bit/8-byte`):
```
 private static final VectorSpecies<Float> F_SPECIES =
            VectorSpecies.of(float.class, VectorShape.forBitSize(L_SPECIES.vectorBitSize() / 2));
```
We added the this requirement in case people run the tests with "-XX:MaxVectorSize=8".

src/hotspot/share/opto/vectorIntrinsics.cpp
```diff
@@ -1382,68 +1382,78 @@ bool LibraryCallKit::inline_vector_mem_masked_operation(bool is_store) {
...
 //
 bool LibraryCallKit::inline_vector_gather_scatter(bool is_scatter) {
   const TypeInstPtr* vector_klass     = gvn().type(argument(0))->isa_instptr();
   const TypeInstPtr* mask_klass       = gvn().type(argument(1))->isa_instptr();
   const TypeInstPtr* elem_klass       = gvn().type(argument(2))->isa_instptr();
   const TypeInt*     vlen             = gvn().type(argument(3))->isa_int();
-  const TypeInstPtr* vector_idx_klass = gvn().type(argument(4))->isa_instptr();
+  const TypeInstPtr* idx_vector_klass = gvn().type(argument(4))->isa_instptr();
+  const TypeInstPtr* idx_elem_klass   = gvn().type(argument(5))->isa_instptr();
+  const TypeLong*    scale_type       = gvn().type(argument(10))->isa_long();
```
XiaohongGong, line 1417: Can we caculate the scale value from the `elem_bt` instead of adding such a constant argument?

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -855,20 +855,44 @@ class LoadVectorNode : public LoadNode {
...
+
+  virtual int Opcode();
+};
+
+class GatherLNode : public GatherNode {
+ public:
+  GatherLNode(Node* c, Node* mem, Node* adr, const TypePtr* at, const TypeVect* vt, Node* indices, Node* scale)
+    : GatherNode(c, mem, adr, at, vt, indices, scale) {
+    assert(indices->bottom_type()->is_vect()->element_basic_type() == T_LONG, "");
+  }
```
XiaohongGong: We can also get the index type info from the index vector. So is it possible that we use a single node for the gather op, and do the different codegen in ad files based on the index element type? The same to the scatter op.

merykitty: I'm not sure that all architectures will support 32-bit indices and 64-bit indices similarly. So I think split them make it easier querying for backend support. In x86 a 64-bit index gather support depends on the vector length while 32-bit one depends on the vector length in byte.

src/hotspot/share/opto/vectorIntrinsics.cpp
merykitty: This is a provision for support of memory segment access where offsets are not scaled.

test/hotspot/jtreg/compiler/vectorapi/VectorGatherScatterTest.java
merykitty: Got it let me add that back

src/hotspot/share/opto/vectorIntrinsics.cpp
XiaohongGong: Got it. Thanks!

src/hotspot/share/opto/vectornode.hpp
XiaohongGong: OK. Make sense to me. And an alternative way is to adding another matcher method to finish the check. But the change sames also complex. So maybe we could rename the node to `LoadGatherIndexINode/LoadGatherIndexLNode` ?

theRealELiu:
> I am not confident in implementing the AArch64 port of this patch, so it would be appreciated if someone can help me with it, thanks very much in advance.

We'd like to take over AArch64 part. Thanks.

PaulSandoz:
> I'm wondering whether it's better to also keep the original array APIs as well?

My recommendation is to shrink the API down to the key primitives and then see if should build things back up with helper methods or with documentation. I suspect we don't need so many methods for this use-case.

A useful performance test could be to operate on the same arguments as the original array APIs implemented using the mapped index vector.

Paul.

merykitty:
@PaulSandoz 

- I have tried adding operations on `MemorySegment` but encountered failures possibly due to mismatched element types. I have not taken a closer look but notably currently our loads/stores with respect to `MemorySegment` only work on native segments of heap segments backed by byte arrays.

- It turns out that adding an offset complicates the range check quite a lot as we have to take into consideration overflow of addition/subtraction. The users can easily simulate this operation by adding/subtracting from the index vector, albeit without the guarantees in cases of overflow, but I think in reality application code does not require such strong guarantees as the JDK itself, maybe a separate helper method to perform strict range checks would be desirable.

PaulSandoz:
> * have tried adding operations on `MemorySegment` but encountered failures possibly due to mismatched element types. I have not taken a closer look but notably currently our loads/stores with respect to `MemorySegment` only work on native segments of heap segments backed by byte arrays.
> 

Ah, now i recall being extra conservative because at the time the alignment constraints of heap-backed segments were also conservative (and the specification was somewhat of a moving target). It's much more flexible now, so it's possible to _scribble_ on any kind of heap-backed segment with unaligned access for values of a different size to the element size.

In ScopedMemoryAccess there is this:
```java
        // @@@ Smarter alignment checking if accessing heap segment backing non-byte[] array
        if (msp.maxAlignMask() > 1) {
            throw new IllegalArgumentException();
        }
```

So i think we can remove this limitation, removing the exception from the segment access JavaDoc and tweaking the doc to use `ValueLayout.JAVA_E_UNALIGNED` rather than `ValueLayout.JAVA_E.withBitAlignment(8)`

And, if needed, tweak the HotSpot internals to deal with the element mismatch.


> * It turns out that adding an offset complicates the range check quite a lot as we have to take into consideration overflow of addition/subtraction. The users can easily simulate this operation by adding/subtracting from the index vector, albeit without the guarantees in cases of overflow, but I think in reality application code does not require such strong guarantees as the JDK itself, maybe a separate helper method to perform strict range checks would be desirable.

Agreed, there is flexibility in allowing the user to control the offset calculation.

